PR #1, opened 2018-04-26
Openeuropa 336: Implement webtools integration

aritomelo:
Develop a module that implements the webtools analytics 

Actions:
 - Custom event, Services, Entity and tests.
 -  Event "WebtoolsImportSettingsEvent"
 -  Event subscriber "WebtoolsImportSettingsSubscriber"
 -  An Entity class called WebtoolsAnalyticsSearch to represent search field on event.
 -  Functional and kernel test


LICENCE.txt
```diff
@@ -0,0 +1,190 @@
+EUROPEAN UNION PUBLIC LICENCE v. 1.2
```
sandervd: It is more future proof to point to the EUPL without version number, instead of including it:
See https://webgate.ec.europa.eu/CITnet/jira/browse/OPENEUROPA-547

behat.yml.dist
```diff
@@ -0,0 +1,23 @@
+default:
+  suites:
+    default:
+      paths:
+        - %paths.base%/tests/features
+      contexts:
+        - Drupal\DrupalExtension\Context\MinkContext
+        - Drupal\DrupalExtension\Context\DrupalContext
+        - Drupal\DrupalExtension\Context\MarkupContext
+        - Drupal\Tests\oe_webtools\Behat\DrupalContext
```
sandervd: - Drupal\DrupalExtension\Context\MarkupContext
 - Drupal\Tests\oe_webtools\Behat\DrupalContext
These classes don't seem to exist

oe_webtools_analytics/oe_webtools_analytics.behat.inc
```diff
@@ -0,0 +1,108 @@
...
+use Drupal\DrupalExtension\Context\DrupalSubContextBase;
+use Drupal\oe_webtools_analytics\Utils\WebtoolsAnalyticsIndex;
+
+/**
+ * Behat step definitions for testing the OE Webtools Analytics module.
+ */
+class OEWebtoolsAnalyticsSubContext extends DrupalSubContextBase {
+
+  /**
+   * Setup demo site.
```
sandervd: Better: install module (site is already installed)

oe_webtools_analytics/oe_webtools_analytics.behat.inc
```diff
@@ -0,0 +1,108 @@
...
+   * @BeforeScenario @webtoolsAnalytics
+   */
+  public function setupWebtoolsAnalytics(BeforeScenarioScope $scope) {
+    \Drupal::service('module_installer')->install(['oe_webtools_analytics']);
+  }
+
+  /**
+   * Checking if the settings on the page match the value on config.
+   *
+   * @Then the rendered script should contains the :arg1 with the same value as the configuration set in settings file
```
sandervd: Better pass in the expected value as an argument, so that we can make real assertions.
In a test it is better to be as explicit (hard-coded) as possible.
I would propose a signature like
Then the analytics script should have (the ):property set to :value
(where :property would be SiteID and :value 12345)

oe_webtools_analytics/oe_webtools_analytics.behat.inc
```diff
@@ -0,0 +1,108 @@
...
+   */
+  public function theSiteIdRenderedOnThePageShouldBeTheSameAsInSettingsFile($siteIdLabel) {
```
sandervd: The implementation doesn't match step definition

oe_webtools_analytics/oe_webtools_analytics.behat.inc
```diff
@@ -0,0 +1,108 @@
...
+    // Message to be displayed in case of failure.
+    $failMsg = "The siteID on settings is $settingsSiteId which differ to the rendered siteID = $onPageSiteId";
+
+    PHPUnit_Framework_Assert::assertEquals($onPageSiteId, $settingsSiteId, $failMsg);
+  }
+
+  /**
+   * Set a new config value for siteID.
+   *
+   * @Then I set the SiteId with :arg1 in settings file
```
sandervd: Then I set the Webtools Analytics SiteID to :siteId

oe_webtools_analytics/oe_webtools_analytics.behat.inc
```diff
@@ -0,0 +1,108 @@
...
+   */
+  public function setSideIdInSettingsFile($siteIdValue) {
+    \Drupal::configFactory()->getEditable('oe_webtools.analytics')
```
sandervd: This is storing the config, but not setting it in the settingsfile. Better update the function name to reflect the implementation.

oe_webtools_analytics/oe_webtools_analytics.behat.inc
```diff
@@ -0,0 +1,108 @@
...
+   * @return array
+   *   An array of  settings found in current page.
+   */
+  private function getAnalyticsSettingsFromPage() : array {
+    $session = $this->getSession();
+    // Inspect if the <script type="application/json">{"utility":"piwik","siteID":"5454544545454","sitePath":["localhost\/open_europa\/oe_webtools\/build\/"]}</script>.
+    $elem = $session->getPage()->find('css', "script[type='application/json']");
+    $settings_list = json_decode($elem->getText(), TRUE);
+
+    $valid_settings = [
```
sandervd: $required_settings

oe_webtools_analytics/oe_webtools_analytics.module
```diff
@@ -0,0 +1,34 @@
+<?php
+
+/**
+ * @file
+ * Contains Drupal hooks.
```
sandervd: Better remove this as it doesn't add much

oe_webtools_analytics/oe_webtools_analytics.module
```diff
@@ -0,0 +1,34 @@
...
+
+use Drupal\oe_webtools_analytics\Event\WebtoolsImportSettingsEvent;
+
+/**
+ * Implements hook_page_attachments().
+ *
+ * Collects the Analytics settings and sets them as a JSON snippet.
+ */
+function oe_webtools_analytics_page_attachments(array &$attachments) {
+  $import_settings_event = new WebtoolsImportSettingsEvent();
```
sandervd: WebtoolsImportSettingsEvent -> AnalyticsSettings ?

oe_webtools_analytics/oe_webtools_analytics.services.yml
```diff
@@ -0,0 +1,10 @@
+services:
+  oe_webtools_analytics.event_subscriber:
+    class: Drupal\oe_webtools_analytics\EventSubscriber\WebtoolsImportSettingsSubscriber
+    arguments: ['@config.factory', '@request_stack']
+    tags:
+      - { name: event_subscriber }
+
+  oe_webtools_analytics.search:
+    class:  Drupal\oe_webtools_analytics\Entity\WebtoolsAnalyticsSearch
```
sandervd: This should most probably not be a service (could be considered to remove this class entirely)

oe_webtools_analytics/src/Controller/WebtoolsController.php
```diff
@@ -0,0 +1,14 @@
...
+
+declare(strict_types = 1);
+
+/**
+ * Controller.
+ */
+
+/**
+ * Class WebtoolsController.
+ */
```
sandervd: This controller can be removed.

oe_webtools_analytics/src/Entity/WebtoolsAnalyticsSearch.php
```diff
@@ -0,0 +1,99 @@
...
+
+declare(strict_types = 1);
+
+/**
+ * Contains the class that will represent the search field on Event listener.
+ *
+ * @see https://webgate.ec.europa.eu/fpfis/wikis/pages/viewpage.action?spaceKey=webtools&title=Piwik
+ */
+
+namespace Drupal\oe_webtools_analytics\Entity;
```
sandervd: This should not be in the entity namespace as it is not an entity.

oe_webtools_analytics/src/Event/WebtoolsImportSettingsEvent.php
```diff
@@ -0,0 +1,217 @@
+<?php
+
+declare(strict_types = 1);
+
+/**
+ * Webtools WebtoolsImportSettingsEvent Event.
```
sandervd: I would change the name: This doesn't import any settings.
Perhaps WebtoolsAnalytics ?

oe_webtools_analytics/src/Utils/WebtoolsAnalyticsIndex.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+namespace Drupal\oe_webtools_analytics\Utils;
+
+/**
+ * Class WebtoolsAnalyticsSearch.
+ *
+ * @package Drupal\oe_webtools_analytics\Entity
+ */
+class WebtoolsAnalyticsIndex {
```
sandervd: Why not move these properties to the event?
There are some instances where they are not used yet (jsonSerialize of the Even class)

oe_webtools_analytics/tests/src/Functional/WebtoolsAnalyticsTest.php
```diff
@@ -0,0 +1,33 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  protected $profile = 'minimal';
+
+  /**
+   * Test the output of the example page.
+   */
+  public function testWebtoolsAnalytics() {
```
sandervd: As this doesn't test anything, I would remove it.

oe_webtools_analytics/tests/src/Kernel/WebtoolsAnalyticsSearchServiceTest.php
```diff
@@ -0,0 +1,30 @@
...
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_webtools_analytics\Kernel;
+
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\oe_webtools_analytics\Entity\WebtoolsAnalyticsSearch;
+
+/**
+ * Test to ensure 'WebtoolsAnalyticsSearch' service is reachable.
```
sandervd: I would remove this test, WebtoolsAnalyticsSearch shouldn't be a service (it is a value-object class for the event), and testing that it exists doesn't add much.

oe_webtools_analytics/tests/src/Kernel/WebtoolsImportSettingsServiceTest.php
```diff
@@ -0,0 +1,30 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public static $modules = ['oe_webtools_analytics'];
+
+  /**
+   * Test for existence of 'WebtoolsAnalyticsSearch' service.
+   */
+  public function testWebtoolsImportSettingsService() {
+    $subscriber = $this->container->get('oe_webtools_analytics.event_subscriber');
```
sandervd: If you want to kernel test this I would propose to make a test module that you enable during the test, that implements the event, and sets some values. Then you can assert the values.

tests/Behat/DrupalContext.php
```diff
@@ -0,0 +1,94 @@
...
+   * | blog     |
+   * | book     |
+   *
+   * @param \Behat\Gherkin\Node\TableNode $modules_table
+   *   The table listing modules.
+   *
+   * @Given the/these module/modules is/are not enabled
+   */
+  public function disableModule(TableNode $modules_table) {
+    // @todo
```
sandervd: Either remove the step, or throw an Exception here, so that it is easy to catch.

tests/Behat/DrupalContext.php
```diff
@@ -0,0 +1,94 @@
...
+   * Provide modules data in the following format:
+   *
+   * | modules  |
+   * | blog     |
+   * | book     |
+   *
+   * @param \Behat\Gherkin\Node\TableNode $modules_table
+   *   The table listing modules.
+   *
+   * @Given the/these module/modules is/are enabled
```
sandervd: Step definition is wrong, as this doesn't enable the modules. Better:
Then I expect these modules to be enabled.

tests/features/webtoolsAnalytics.feature
```diff
@@ -0,0 +1,30 @@
+@api @webtoolsAnalytics
+Feature: Check Piwik
+  In order to check if the type attribute is set for the Piwik element.
```
sandervd: In order to have an insight in the usage of my website
As a site owner
I want to run analytics on my site

tests/features/webtoolsAnalytics.feature
```diff
@@ -0,0 +1,30 @@
...
+  As a developer and or a site administrator
+  I want to check Piwik is available.
+
+  Background:
+    Given these modules are enabled
+      | modules                            |
+      | oe_webtools_analytics              |
+    And I am logged in as a user with the "administrator" role
+
+  Scenario: Check if the PIWIK script is embedded into the page correctly
```
sandervd: > Scenario: Check if the PIWIK script is embedded into the page correctly

 Scenario: Check if the Analytics configuration is embedded in the page

tests/features/webtoolsAnalytics.feature
```diff
@@ -0,0 +1,30 @@
...
+    Given I am on the homepage
+    Then the response status code should be 200
+    Then the rendered script should contains the "siteID" with the same value as the configuration set in settings file
```
sandervd: See previous comment:
Then the analytics script should have the SiteID set to 12345

oe_webtools_analytics/oe_webtools_analytics.behat.inc
ademarco: We will have the following issue when using Drupal Extension Behat context autoloading (i.e. when contexts are defined in a `.behat.inc`): if `OEWebtoolsAnalyticsSubContext` uses a trait or a class that is autoloaded via the `autoload-dev` property of the module's `composer.json` such trait or class will not be available when the `oe_webtools` module will be used as a dependency in another project; at the same time the Drupal Extension will include the class anyway, which will result in a fatal.

Since we use a Composer-based workflow we don't need third-party autoloading methods, we can just use Composer `autoload` and `autoload-dev` properties.

tests/Behat/DrupalContext.php
aritomelo: Here it is checking if the module is enabled, on setUp it is where the module has been enabled.

tests/features/webtoolsAnalytics.feature
aritomelo: But we don't have a specific requirement for the site owner, so I would suggest to let it cover wider user range at the moment, then when we know about the roles and permission we can restrict it.

sandervd: Ah, well I was trying to rephrase it reflect more business value. The site owner is not a role in the sense of a Drupal role.

tests/features/webtoolsAnalytics.feature
sandervd: I wouldn't override it in the settings file, except during development of the analytics code, or in production. While running the CI we can leave it as-is.

composer.json
```diff
@@ -0,0 +1,66 @@
+{
+    "name": "openeuropa/oe_webtools",
+    "description": "OpenEuropa Drupal module template.",
```
drupol: Missing module description.

oe_webtools.info.yml
```diff
@@ -0,0 +1,5 @@
+name: Webtools Integration
+description: OpenEuropa webtools integration.
+package: OpenEuropa Webtools
```
drupol: Are we sure about the package here ? Shouldn't it be `OpenEuropa` only ?

oe_webtools.libraries.yml
```diff
@@ -0,0 +1,5 @@
+# Third-party library (CDN).
+drupal.webtools-smartloader:
```
drupol: Shouldn't we prepend the library name with the name of the module ?

oe_webtools_analytics/oe_webtools_analytics.info.yml
```diff
@@ -0,0 +1,8 @@
+name: OpenEuropa Webtools Analytics
+description: Provides the configuration settings for webtools analytics integration.
+package: OpenEuropa Webtools
```
drupol: I guess the package name must stay "OpenEuropa Webtools" as this module has submodules ?
If this is the case, then my previous comment about the package can be discarded.

oe_webtools_analytics/src/AnalyticsEventInterface.php
```diff
@@ -0,0 +1,214 @@
...
+
+  /**
+   * The site unique identifier.
+   */
+  const SITE_ID = 'siteID';
+
+  /**
+   * Representing the 403 key in settings.
+   */
+  const  IS403 = 'is403';
```
drupol: Two spaces instead of one. Make it consistent (see line 20)

oe_webtools_analytics/src/Controller/WebtoolsController.php
```diff
@@ -0,0 +1,14 @@
+<?php
```
drupol: Why do we keep this empty file ?

oe_webtools_analytics/src/Entity/SearchParameters.php
```diff
@@ -0,0 +1,105 @@
...
+ */
+
+namespace Drupal\oe_webtools_analytics\Entity;
+
+use JsonSerializable;
+
+/**
+ * Class SearchParameters.
+ *
+ * @package Drupal\oe_webtools_analytics\Entity
```
drupol: The @package key can now be removed I think.

oe_webtools_analytics/src/Entity/SearchParameters.php
```diff
@@ -0,0 +1,105 @@
...
+   * {@inheritdoc}
+   */
+  public function setKeyword(string $keyword): void {
+    $this->keyword = $keyword;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setCategory($category): void {
```
drupol: Can we add the type of `$category` as well ?

oe_webtools_analytics/src/Entity/SearchParametersInterface.php
```diff
@@ -0,0 +1,76 @@
...
+   */
+  public function setKeyword(string $keyword): void;
+
+  /**
+   * Sets search category.
+   *
+   * @param string $category
+   *   An optional string value.
+   */
+  public function setCategory($category): void;
```
drupol: Can we add the `$category` type ?

oe_webtools_analytics/src/Entity/SearchParameters.php
```diff
@@ -0,0 +1,105 @@
...
+   */
+  public function isSetKeyword() {
+    return !empty($this->keyword);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function jsonSerialize() {
+    $search = [
```
drupol: Maybe we can use `get_object_vars($this);` for this ?

oe_webtools_analytics/src/Event/AnalyticsEvent.php
```diff
@@ -0,0 +1,277 @@
...
+   */
+  public function getUtility(): string {
+    return $this->utility;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function jsonSerialize() {
+    $data = [
```
drupol: `get_object_vars($this)` might help here ?

oe_webtools_analytics/src/Event/AnalyticsEvent.php
```diff
@@ -0,0 +1,277 @@
...
+  public function __toString() {
+    return json_encode($this);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function isValid() : bool {
+    // SiteId is required.
+    if (!$this->siteId) {
```
drupol: I think for a better readability it would be better to use:

`return !empty($this->siteId);`

oe_webtools_analytics/src/EventSubscriber/AnalyticsEventSubscriber.php
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * AnalyticsEventSubscriber constructor.
+   *
+   * @param \Drupal\Core\Config\ConfigFactoryInterface $configFactory
+   *   The configuration object.
+   * @param \Symfony\Component\HttpFoundation\RequestStack $requestStack
+   *   The request on the stack.
+   */
+  public function __construct(ConfigFactoryInterface $configFactory, RequestStack $requestStack) {
+    // Get id from settings.php!
```
drupol: Instead of using if/else, wouldn't it be better to, first check for the availability of the settings, then react accordingly ?
This would reduce overall code complexity.

```php
if (settings_not_available) {
  throw Exception();
}

// rest of the code here.
```

oe_webtools_analytics/src/EventSubscriber/AnalyticsEventSubscriber.php
```diff
@@ -0,0 +1,100 @@
...
+   */
+  public function onSetSiteDefaults(AnalyticsEventInterface $event) {
+    // SiteID.
+    $site_id = $this->config->get(AnalyticsEventInterface::SITE_ID);
+    if ($site_id) {
+      $event->setSiteId((string) $site_id);
+    }
+
+    // SitePath.
+    if (\Drupal::configFactory()->get('oe_webtools.analytics')->get(AnalyticsEventInterface::SITE_PATH)) {
```
drupol: I think we can inject this instead of using `\Drupal` ? No ?

LICENCE.txt
```diff
@@ -0,0 +1,4 @@
+EUROPEAN UNION PUBLIC LICENCE
```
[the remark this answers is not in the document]
aritomelo: This was suggested by @sandervd as the outcome from  https://webgate.ec.europa.eu/CITnet/jira/browse/OPENEUROPA-547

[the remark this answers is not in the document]
drupol: Ok, I'm perfectly fine with this, then, let's use that in all projects then. @ademarco ?

oe_webtools.info.yml
aritomelo: No requirement for that and even recommendation, so it is open for discussion!

oe_webtools_analytics/src/Entity/SearchParameters.php
aritomelo: As it implements the jsonserializable class it is mandatory to have it.

drupol: Sure sure, but we should use `get_object_vars($this);` in it :-)

aritomelo: In this case is not possible once the class members are private, so not available through the suggestion method.

oe_webtools_analytics/src/EventSubscriber/AnalyticsEventSubscriber.php
aritomelo: Cool!

oe_webtools_analytics/src/Entity/SearchParameters.php
drupol: Indeed :+1:

oe_webtools_analytics/src/EventSubscriber/AnalyticsEventSubscriber.php
```diff
@@ -46,13 +46,11 @@ class AnalyticsEventSubscriber implements EventSubscriberInterface {
...
   public function __construct(ConfigFactoryInterface $configFactory, RequestStack $requestStack) {
     // Get id from settings.php!
-    if ($configFactory->get('oe_webtools.analytics')) {
-      $this->config = $configFactory->get('oe_webtools.analytics');
-      $this->requestStack = $requestStack;
-    }
-    else {
+    if (!$configFactory->get('oe_webtools.analytics')) {
       throw new \InvalidArgumentException(t('The "oe_webtools.analytics" settings is missing from settings.php', [], ['context' => 'oe_webtools']));
     }
```
drupol: Usually we had an empty line after the `if` for better readability.

oe_webtools_analytics/src/EventSubscriber/AnalyticsEventSubscriber.php
```diff
@@ -63,26 +61,22 @@ public function __construct(ConfigFactoryInterface $configFactory, RequestStack
...
 
     // Set exception flags when access is denied, or page not found.
     $request_exception = $this->requestStack->getCurrentRequest()->attributes->get('exception');
     if ($request_exception instanceof NotFoundHttpException) {
       $event->setIs404Page();
     }
-
-    if ($request_exception instanceof AccessDeniedHttpException) {
+    elseif ($request_exception instanceof AccessDeniedHttpException) {
```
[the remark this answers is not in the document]
drupol: Ok I shut up.

drupol:
Thanks for the changes, I just approved.

oe_webtools_analytics/src/EventSubscriber/AnalyticsEventSubscriber.php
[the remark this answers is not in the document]
aritomelo: Thx @drupol !

oe_webtools_analytics/src/EventSubscriber/AnalyticsEventSubscriber.php
```diff
@@ -0,0 +1,94 @@
...
+   * AnalyticsEventSubscriber constructor.
+   *
+   * @param \Drupal\Core\Config\ConfigFactoryInterface $configFactory
+   *   The configuration object.
+   * @param \Symfony\Component\HttpFoundation\RequestStack $requestStack
+   *   The request on the stack.
+   */
+  public function __construct(ConfigFactoryInterface $configFactory, RequestStack $requestStack) {
+    // Get id from settings.php!
+    if (!$configFactory->get('oe_webtools.analytics')) {
```
imanoleguskiza: This statement is always true, the get function always returns a Config object so we will never trigger the exception. You should instead try check against the raw values or against an specific value in the config. On the same note, its strange that here we trigger a full blown exception when the settings are not there but on the requirements hook we just give a warning. Shouldn't we prevent the usage of the module alltogether if the configuration is not in place?

tests/features/webtoolsAnalytics.feature
```diff
@@ -0,0 +1,32 @@
+@api
+Feature: Check Analytics
```
imanoleguskiza: I'm not sure if behat tests are what we need for this module. None of the functionality it provides is based on no UI and requires no input from a site's user so I don't think we could create believable user stories. Instead it seems to me that kernel tests should be more appropriate. cc @pfrenssen

oe_webtools.libraries.yml
aritomelo: @ademarco @sandervd What do you think about this topic?

oe_webtools_analytics/oe_webtools_analytics.info.yml
aritomelo: Yes, it has!

oe_webtools_analytics/src/Event/AnalyticsEvent.php
aritomelo: No, the members class are private!

oe_webtools_analytics/src/EventSubscriber/AnalyticsEventSubscriber.php
aritomelo: Sorry but I don't agree and the Drupal CS does not mention it!

drupol: I know but I don't agree with the "Drupal CS does not mention it".

It's true that nothing "regulates" the usage of empty lines and also do not consider Drupal CS to be the holy grail, it has many inconsistencies.

Usually I add spaces to clearly see the different condition blocks. But I'm not the only one.
Famous libraries are doing it "by default" (see Geocoder: https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/GoogleMaps/GoogleMaps.php#L105) or PHPunit (https://github.com/sebastianbergmann/phpunit/blob/master/src/Util/ErrorHandler.php) (there are many examples in the code)

They are also adding an empty line before the `return` statement.

Now, of course this is not a blocker at all, it's up to you to decide to follow this or not.

oe_webtools_analytics/oe_webtools_analytics.install
```diff
@@ -0,0 +1,25 @@
...
+
+declare(strict_types = 1);
+
+/**
+ * Implements hook_requirements().
+ */
+function oe_webtools_analytics_requirements($phase) {
+  $requirements = [];
+  // Report missing siteId on settings.
+  if (\Drupal::configFactory()->get('oe_webtools.analytics')->get('site_id')) {
```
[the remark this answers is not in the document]
aritomelo: Well spotted!

tests/features/webtoolsAnalytics.feature
aritomelo: Because is about rendering element the Behat seems a valid way to test it.

oe_webtools_analytics/oe_webtools_analytics.behat.inc
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * Create the analytics settings on runtime.
+   *
+   * @param string $property
+   *   The property label, eg.: siteID, sitePath, is403, is404.
+   * @param string $value
+   *   The actual property value.
+   *
+   * @Then I should be able to set the property :arg1 with the value :arg2
```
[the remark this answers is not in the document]
aritomelo: I do agree with you, however, this is a result of the discussion between me and @sandervd where we decided that the scenario should create the setting and then test if the markup was correctly in place.

oe_webtools_analytics/src/EventSubscriber/AnalyticsEventSubscriber.php
aritomelo: I would suggest we bring this discussion to the retro and we set up the CS we are going to follow.

As you mention there are several libraries using it, that's true, but I can say that there are several core Drupal modules not using it (for example node module), so this shows that we need to have a team agreement for this kind of matters!
cc: @ademarco @voidtek @imanoleguskiza @sandervd

drupol: Mmhm.

As it's not a blocker, I don't think we should spend time "once again" in retro and/or any meeting for dealing with this.
And as it's not in a CS it's up to you to decide to do it or not.

We must think about being flexible and we of course can agree to disagree without doing meetings...

oe_webtools.libraries.yml
sandervd: :man_shrugging:

aritomelo:
@drupol Good work!

aritomelo:
@imanoleguskiza Regarding your comment about the use of "event/event dispacher", this is the most use approach to allow others module to hook a custom module in D8, so it is antecipating the case where modules need to hook it!

PR #6, opened 2018-07-17
OPENEUROPA-721: Update code-review and grump.yml.dist file.

imanoleguskiza:
### Description

Update code-review to 0.3 and update grumphp.dist.yml so that we don't override default tasks

### Change log
 - Changed: Code review to version 0.3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     "drupal/console": "^1.6",
     "drush/drush": "^9",
     "nikic/php-parser": "~3",
-    "openeuropa/code-review": "^0.3",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -17,7 +17,7 @@
     "drupal/console": "^1.6",
     "drush/drush": "^9",
     "nikic/php-parser": "~3",
-    "openeuropa/code-review": "^0.2",
+    "openeuropa/code-review": "^0.3",
```
pfrenssen: Always use the tilde for dependencies that are on version 0, so that it is forwards compatible. This would be better as `~0.3`.

PR #8, opened 2018-08-02
OPENEUROPA-1108: Module to send the availability of the language for one content

upchuk:
### Description

* Created the submodule
* Info on how it works and how to test are in the README.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_webtools_laco_service/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+A Webtools Laco service that provides information on whether entity has a translation in a specific language.
+
+### How it works
+
+This module exposes a new route for each entity type that "mirrors" the canonical route of entities. These routes have 
+the same paths as the canonical entity routes but use different requirements for matching. If these requirements
+are matched in the request, the module will return Laco information for that entity.
+
+### How to test
```
imanoleguskiza: I'm not sure if we need to add this section to the Readme but I won't block the PR because of it.

modules/oe_webtools_laco_service/src/StackMiddleware/LacoServiceMiddleware.php
```diff
@@ -0,0 +1,73 @@
...
+   */
+  public function __construct(HttpKernelInterface $http_kernel) {
+    $this->httpKernel = $http_kernel;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function handle(Request $request, $type = self::MASTER_REQUEST, $catch = TRUE) {
+    if (!$this->isLacoServiceRequest($request)) {
```
brummbar: This whole method can be changed to
```
if ($this->isLacoServiceRequest($request)) {
  $request->attributes->set('_format', 'laco');
}

return $this->httpKernel->handle($request, $type, $catch);
```

modules/oe_webtools_laco_service/README.md
```diff
@@ -0,0 +1,35 @@
+# OpenEuropa Webtools Laco Service
+
+A Webtools Laco service that provides information on whether entity has a translation in a specific language.
```
ademarco: The LACO service should also work for non-entity pages, in which case we simply return all available site languages.

In Drupal 7 this feature was covered with [this test](https://github.com/ec-europa/platform-dev/blob/release-2.5/profiles/common/modules/custom/nexteuropa_laco/tests/LanguageCoverageServiceTest.php#L18).

modules/oe_webtools_laco_service/src/StackMiddleware/LacoServiceMiddleware.php
upchuk: Done, please check.

upchuk:
@ademarco As discussed, we pause the debug issue.

Implemented the non-entity page scenario.

pfrenssen:
I am wondering if `oe_webtools_laco_service` is the right name for the module, since this module does not actually provide the service but integrates with it. Wouldn't it be better to just call it `oe_webtools_laco` to be consistent with the other modules that integrate with Webtools?

upchuk:
@pfrenssen The module doesnt integrate with a service, it provides one (the data for laco). In parallel, we have another module that prints out the widget (oe_webtools_laco_widget) which the webtools service uses to make requests for data we provide in oe_webtools_laco_service :)

modules/oe_webtools_laco_service/src/LacoServiceHeaders.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_webtools_laco_service;
+
+/**
+ * A simple class that contains the headers used by the Laco service.
+ */
+class LacoServiceHeaders {
```
brummbar: This is a good use case to declare the class as `final`. It's a common thing when classes only contain constants.

modules/oe_webtools_laco_service/src/StackMiddleware/LacoServiceMiddleware.php
```diff
@@ -0,0 +1,70 @@
...
+   */
+  private function isLacoServiceRequest(Request $request) {
+    if ($request->getMethod() !== 'HEAD') {
+      return FALSE;
+    }
+
+    $headers = $request->headers;
+    $header = LacoServiceHeaders::HTTP_HEADER_SERVICE_NAME;
+    $value = LacoServiceHeaders::HTTP_HEADER_SERVICE_VALUE;
+    if ($headers->get($header) === $value && $headers->has(LacoServiceHeaders::HTTP_HEADER_LANGUAGE_NAME)) {
```
brummbar: Nit, we could just do the return here.

modules/oe_webtools_laco_service/tests/src/Kernel/LacoServiceTest.php
```diff
@@ -0,0 +1,276 @@
...
+
+  /**
+   * Laco service test for entities.
+   *
+   * Tests that the Laco service works with entities: requests made to entity
+   * routes which contain certain headers and language will return empty
+   * responses that contain a status code that confirms the existence of a
+   * translation in that language.
+   */
+  public function testEntityLacoService() {
```
brummbar: This class is missing return types.

upchuk:
@brummbar @ademarco @sandervd 

Alright, pushed in two commits: one to prove the test failing and one to prove the request policy for caching actually works.

It wasn't easy but managed to make this testable with KernelTest instead of BrowserTest. It required some creativity (see test module enabled for it) which wrangled a bit the core Page Cache module services. But in the end, it was minimal code.

PR #9, opened 2018-08-03
OPENEUROPA-1107: Include the widget of webtools for Laco

richardcanoe:
### Description

Includes also the runner change to install minimal instead of standard profile.


modules/oe_webtools_laco_widget/oe_webtools_laco_widget.module
```diff
@@ -0,0 +1,54 @@
...
+ */
+function oe_webtools_laco_widget_page_attachments(array &$attachments) {
+  /** @var \Symfony\Component\EventDispatcher\EventDispatcherInterface $dispatcher */
+  $dispatcher = \Drupal::service('event_dispatcher');
+
+  $include = ['#page-wrapper'];
+  $exclude = ['.nolaco', '.more-link', '.pager'];
+  $event = new LacoWidgetConfig($include, $exclude);
+  /** @var \Drupal\oe_webtools_laco_widget\Event\LacoWidgetConfig $event */
+  $event = $dispatcher->dispatch(LacoWidgetConfig::LACO_WIDGET_CONFIG_EVENT, $event);
```
imanoleguskiza: I don't understand why we use an event to configure this functionality. Doesn't it make more sense to simply use drupal configuration? That way the default configuration can be exported like all the rest and can be easily changed through a form for example.

upchuk: @imanoleguskiza We are using an event so that we can allow implementing sites to alter the classes. But this is not something for editors for example, only devs should be allowed to change this, which means anyway code. And it means also unnecessary form for configuring, config export and stuff.

imanoleguskiza: I'm not so sure this will be the work of devs to configure. I'm quite certain webmasters and people from CEM will need to make adjustements to this quite often (at least I believe this was the case on NE sites). Now it may be the case that we are moving away from this approach but I couldn't see this reflected in the tickets so far.

upchuk: @imanoleguskiza Checked on how it's used. Atm it's a simple bean block dump if the json and placed in the footer. So no configurability for now.

imanoleguskiza: Yes it's a bean block but the contents of the json are defined by whoever creates the bean, so whoever has the appropriate permission can define what is included and what is excluded. I believe that we are losing functionality here but the base is there so I'll approve for now and follow up with discussion later.

upchuk:
@ademarco Done, check it out please.

ademarco:
It's looking good, shall we maybe add a section on the README.md on how to configure it for now? So overriding the staged configuration on `settings.php`. After that we can merge it and close it.

upchuk:
@ademarco Done.

modules/oe_webtools_laco_widget/README.md
```diff
@@ -0,0 +1,51 @@
+# OpenEuropa Webtools Laco Widget
+
+Provides a widget that integrates with the Webtools Laco service and which checks if a 
+link to o page or document is available in other translations.
```
ademarco: Typo: `to o page`

modules/oe_webtools_laco_widget/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+
+### Configuration
+
+The default widget that is being output is the following:
+
+```
+{
+    "service"    : "laco",
+    "include"    : "#page-wrapper",
```
ademarco: Use `body` so it can work by default anywhere regardless of which theme we are using.

modules/oe_webtools_laco_widget/config/install/oe_webtools_laco_widget.settings.yml
```diff
@@ -0,0 +1,6 @@
+include: ['#page-wrapper']
```
ademarco: Use `body` as `include` default.

modules/oe_webtools_laco_widget/oe_webtools_laco_widget.module
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Implements hook_page_attachments().
+ */
+function oe_webtools_laco_widget_page_attachments(array &$attachments) {
+  $logger = \Drupal::logger('oe_webtools_laco_widget');
+  $config = \Drupal::config('oe_webtools_laco_widget.settings');
+
+  $include = $config->get('include');
+  if (!$include) {
+    $logger->info('There are no selectors to include the Laco widget for. Widget is not being printed.');
```
ademarco: The purpose of this message is to alert the site developer that the module should not be enabled at all if no include is specified, shall we add this as a suggestion to the message? We could also promote this to a warning.

PR #10, opened 2018-08-17
OPENEUROPA-1165: Create a docker-compose.yml.dist file on oe_webtools

upchuk:
### Description

* Created the `docker-compose.yml.dist` file with the existing defaults and some extras for Mac users commented out. Added also mysql port mapping to be able to access it from outside the container.
* Gitignored the `docker-compose.yml` file.
* Updated the readme.

### Change log

- Added: 
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
no commands

```


README.md
```diff
@@ -43,6 +43,11 @@ Requirements:
 - [Docker](https://www.docker.com/get-docker)
 - [Docker-compose](https://docs.docker.com/compose/)
 
+Copy `docker-compose.yml.dist` into `docker-compose.yml`.
+
+You can make any alterations you need for your local Docker setup but the defaults
+should be enough to set the site up.
```
netlooker: Instead of the 'site', I would suggest using the word 'project' as it might give a wrong impression to some users. Also, I think that in this case there should be a colon before the but because of connecting two independent clauses ;)

docker-compose.yml.dist
```diff
@@ -0,0 +1,32 @@
...
+      # XDEBUG_CONFIG: "remote_enable=1 remote_host=10.254.254.254 remote_port=9000 idekey=PHPSTORM remote_autostart=1" # Mac users. Run this command in the terminal to map the IP address: sudo ifconfig en0 alias 10.254.254.254 255.255.255.0
+      PHP_IDE_CONFIG: "serverName=Docker"
+  mysql:
+    image: percona/percona-server:5.6
+    environment:
+      MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
+    ports:
+      - 3306:3306
+
+#### Mac users: uncomment the "volumes" key to enable the NFS file sharing.
```
netlooker: I'm wondering how to inform users that they must run a specific script beforehand. Maybe we could add an URL to this page https://github.com/openeuropa/openeuropa/blob/master/docs/starting/tooling.md#using-docker-on-macos.

docker-compose.yml.dist
```diff
@@ -0,0 +1,32 @@
...
+    environment:
+      XDEBUG_CONFIG: "remote_enable=1 remote_host=${DOCKER_HOST_IP} remote_port=9000 idekey=PHPSTORM remote_autostart=1" # Non Mac users.
+      # XDEBUG_CONFIG: "remote_enable=1 remote_host=10.254.254.254 remote_port=9000 idekey=PHPSTORM remote_autostart=1" # Mac users. Run this command in the terminal to map the IP address: sudo ifconfig en0 alias 10.254.254.254 255.255.255.0
+      PHP_IDE_CONFIG: "serverName=Docker"
+  mysql:
+    image: percona/percona-server:5.6
+    environment:
+      MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
+    ports:
+      - 3306:3306
```
ademarco: I'd not set a port as it might conflict with an already running mysql instance on the host machine, shall we simply comment these lines out?

docker-compose.yml.dist
```diff
@@ -0,0 +1,32 @@
...
+    image: fpfis/httpd-php-dev:7.1
+    working_dir: /var/www/html
+    ports:
+      - 8080:8080
+    volumes:
+      - .:/var/www/html # Non Mac users.
+      # - nfsmount:/var/www/html # Mac Users with the nfsmount volume.
+    environment:
+      XDEBUG_CONFIG: "remote_enable=1 remote_host=${DOCKER_HOST_IP} remote_port=9000 idekey=PHPSTORM remote_autostart=1" # Non Mac users.
+      # XDEBUG_CONFIG: "remote_enable=1 remote_host=10.254.254.254 remote_port=9000 idekey=PHPSTORM remote_autostart=1" # Mac users. Run this command in the terminal to map the IP address: sudo ifconfig en0 alias 10.254.254.254 255.255.255.0
```
ademarco: Both mac and non-mac will work with `10.254.254.254` so we can un-comment this line, I'd move the comment above for better readability, but up to you! :)

PR #11, opened 2018-08-21
OPENEUROPA-768: Add oe drupal core require dev.

Medi4tor:
### Description

Have OpenEuropa Webtools use openeuropa/drupal-core-require-dev

### Change log

Updated composer requirements

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,9 +18,9 @@
     "drush/drush": "^9",
     "nikic/php-parser": "~3",
     "openeuropa/code-review": "~0.3",
+    "openeuropa/drupal-core-require-dev": "~8.6@alpha",
     "openeuropa/task-runner": "^0.5",
-    "symfony/dom-crawler": "^3",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -45,11 +45,6 @@
   },
   "extra": {
     "composer-exit-on-patch-failure": true,
```
upchuk: We need an extra line inside `extras`:

```
"enable-patching": true
```

composer.json
```diff
@@ -18,9 +18,9 @@
     "drush/drush": "^9",
     "nikic/php-parser": "~3",
     "openeuropa/code-review": "~0.3",
+    "openeuropa/drupal-core-require-dev": "~8.6@alpha",
```
voidtek: It's not necessary to specify `@alpha` in this dependency
Could you change by:
`"openeuropa/drupal-core-require-dev": "~8.6",`

PR #12, opened 2018-08-22
OPENEUROPA-1126: Use logger factory as a service.

aritomelo:
## OPENEUROPA-1126
 - Use Logger factory as a service
 - Made all report messages from this module to be of type  "oe_webtools"


modules/oe_webtools_analytics/src/EventSubscriber/AnalyticsEventSubscriber.php
```diff
@@ -37,6 +38,13 @@ class AnalyticsEventSubscriber implements EventSubscriberInterface {
    */
   private $requestStack;
 
+  /**
+   * {@inheritdoc}
+   *
+   * @var \Drupal\Core\Logger\
```
upchuk: Incorrect namespace here in the docblock.

modules/oe_webtools_analytics/src/EventSubscriber/AnalyticsEventSubscriber.php
```diff
@@ -51,11 +59,14 @@ class AnalyticsEventSubscriber implements EventSubscriberInterface {
...
    *   The request on the stack.
+   * @param \LoggerChannelFactoryInterface $loggerFactory
+   *   The logger channel factory.
    */
-  public function __construct(ConfigFactoryInterface $configFactory, RequestStack $requestStack) {
+  public function __construct(ConfigFactoryInterface $configFactory, RequestStack $requestStack, LoggerChannelFactoryInterface $loggerFactory) {
     // Get id from settings.php!
     $this->config = $configFactory->get(AnalyticsEventInterface::CONFIG_NAME);
     $this->requestStack = $requestStack;
+    $this->loggerFactory = $loggerFactory;
```
upchuk: I would not set logger factory in the class, but directly the channel.

```
$this->logger = $loggerFactory->get('oe_webtools');
```

And then you have `$logger` everywhere you need to log.

modules/oe_webtools_analytics/src/EventSubscriber/AnalyticsEventSubscriber.php
```diff
@@ -85,13 +96,11 @@ public function getConfig(): ImmutableConfig {
...
-    $factory = \Drupal::service('logger.factory');
-
     // SiteID must exist and be an integer.
     $site_id = $this->getConfig()->get(AnalyticsEventInterface::SITE_ID);
     if (!is_numeric($site_id)) {
-      $factory
-        ->get('default')
+      $this->loggerFactory
+        ->get('oe_webtools')
         ->debug('The setting "' . AnalyticsEventInterface::SITE_ID . '" is missing from settings file.');
```
upchuk: I would make this a `warning` rather than `debug` no?

PR #13, opened 2018-09-04
OPENEUROPA-785: Make sure component dependencies are as relaxed as possible.

aritomelo:
Components dependencies:

  - Update some version to use tilde, so it uses the minimal minor version and also for a consistency use decimal number as the version.

  - Update some packages to use minimal minor version.

  - Added composer boundary to test lowest and highest.

  - Added packagist badge.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,7 @@
         "openeuropa/code-review": "^1.0.0-alpha4",
         "openeuropa/drupal-core-require-dev": "~8.6",
         "openeuropa/task-runner": "~1.0",
+        "phpunit/phpunit": "~6.0",
         "symfony/dom-crawler": "~3.0"
     },
     "scripts": {


composer.json
```diff
@@ -10,17 +10,17 @@
...
     "drupal-composer/drupal-scaffold": "^2.2",
     "drupal/config_devel": "~1.2",
-    "drupal/console": "^1.6",
-    "drush/drush": "^9",
-    "nikic/php-parser": "~3",
+    "drupal/console": "~1.0",
+    "drush/drush": "~9.0",
+    "nikic/php-parser": "~3.0",
     "openeuropa/code-review": "~0.3",
     "openeuropa/drupal-core-require-dev": "~8.6",
```
pfrenssen: Since we are setting the requirement here to `~8.6` I would also apply the same version constraint to `drupal/core` (in the `require` section above). They are supposed to be set to exactly the same version.

pfrenssen:
Not all my remarks were addressed, and new problems have been introduced. It is now also more difficult to review since the spacing of the file has been changed.

Let's change this first back to the original spacing so we can actually see which parts of the files have changed.

drupol:
The spacing must me 4 spaces, i fixed it as well.

drupol:
You can une this to review without the spacing changes : https://github.com/openeuropa/oe_webtools/pull/13/files?utf8=%E2%9C%93&diff=split&w=1

pfrenssen:
Thanks, let's keep the spaces on 4. I can review it now easily.

I fixed some small obvious things but the lowest dependency range is failing, this will be a case of trial and error to find what is causing it :(

PR #15, opened 2018-10-04
OPENEUROPA-1247: Cleanup and minor fixes for Analytics.

dxvargas:
### Description

Improve analytics module of the webtools component.

### Change log

- Fixed: Correctly set the site path without using globals
- Fixed: Incorrect “LoggerChannelInterface” reference
- Changed: Move the SearchParameters and its interface outside the “Entity” folder
- Changed: Inside AnalyticsEvent, switch the condition order inside isValid() to be consistent with Drupal coding standards
- ~Changed: Inside the ConfigurationTest, don’t use drupal_flush_all_caches() and also test the changes in config are reflected in the widget output~
- Removed: Useless “AnalyticsEventSubscriberServiceTest”
- Removed: Useless $search property
- Removed: “getRequestStack” and getConfig methods


composer.json
```diff
@@ -6,21 +6,22 @@
...
-        "drupal/core": "~8.6",
+        "drupal/core": "^8.6",
+        "ext-json": "*",
         "php": "^7.1"
     },
     "require-dev": {
         "composer/installers": "~1.5",
         "cweagans/composer-patches": "~1.6",
-        "drupal-composer/drupal-scaffold": "^2.5.2",
+        "drupal-composer/drupal-scaffold": "~2.5.2",
```
aritomelo: Do we really wanna change this, for what I understand when we have patch version we should use ^, may I wrong?

@pfrenssen What is your thought on this matter?

dxvargas: @aritomelo changing from the "^"to the "~" won't prevent patch versions to be applied. Actually right now the version used will be "2.5.4".
The difference is that it will prevent from changing to "2.6.x". once there is that version. That conforms to our policy to have more strict rules and more stability for dev environment.
If it was a require entry (instead of a require_dev), then we would prefer "^", because it's our policy to keep constraints as open as possible for the community (non dev) users.

PR #17, opened 2018-11-07
OPENEUROPA-665: Add section rules to Webtool Analytics.

imanoleguskiza:
### Description

The Webtools analytics team needs to set the site section based on a regex by simple configuration.

### Change log

- Added: Add config entities for section rules

### DIFF
--- a/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
+++ b/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
@@ -9,12 +9,12 @@ use Drupal\oe_webtools_analytics\AnalyticsEventInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
 /**
- * Event subscriptor for the Webtools Analytics event.
+ * Event subscriber for the Webtools Analytics event.
  */
 class WebtoolsAnalyticsEventSubscriber implements EventSubscriberInterface {
 
   /**
-   * Kernel request event handler.
+   * Webtools Analytics event handler.
    *
    * @param \Drupal\oe_webtools_analytics\AnalyticsEventInterface $event
    *   Response event.
--- a/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
+++ b/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
@@ -24,7 +24,7 @@ class WebtoolsAnalyticsEventSubscriber implements EventSubscriberInterface {
    * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
    *   Thrown if the storage handler couldn't be loaded.
    */
-  public function setSection(AnalyticsEventInterface $event) {
+  public function setSection(AnalyticsEventInterface $event): void {
     $storage = \Drupal::entityTypeManager()->getStorage('webtools_analytics_rule');
     $rules = $storage->loadMultiple();
     $current_uri = \Drupal::request()->getRequestUri();
@@ -39,7 +39,7 @@ class WebtoolsAnalyticsEventSubscriber implements EventSubscriberInterface {
   /**
    * {@inheritdoc}
    */
-  public static function getSubscribedEvents() {
+  public static function getSubscribedEvents(): array {
     // Subscribing to listening to the Analytics event.
     $events[AnalyticsEvent::NAME][] = ['setSection'];
 
--- a/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
+++ b/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
@@ -18,11 +18,11 @@ use Symfony\Component\HttpFoundation\RequestStack;
 class WebtoolsAnalyticsEventSubscriber implements EventSubscriberInterface {
 
   /**
-   * The event type manager.
+   * The entity type manager.
    *
    * @var \Drupal\Core\Entity\EntityTypeManagerInterface
    */
-  private $eventTypeManager;
+  private $entityTypeManager;
 
   /**
    * The request stack.
@@ -34,13 +34,13 @@ class WebtoolsAnalyticsEventSubscriber implements EventSubscriberInterface {
   /**
    * WebtoolsAnalyticsEventSubscriber constructor.
    *
-   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $eventTypeManager
-   *   The event type manager.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entityTypeManager
+   *   The entity type manager.
    * @param \Symfony\Component\HttpFoundation\RequestStack $requestStack
    *   The request stack.
    */
-  public function __construct(EntityTypeManagerInterface $eventTypeManager, RequestStack $requestStack) {
-    $this->eventTypeManager = $eventTypeManager;
+  public function __construct(EntityTypeManagerInterface $entityTypeManager, RequestStack $requestStack) {
+    $this->entityTypeManager = $entityTypeManager;
     $this->requestStack = $requestStack;
   }
 
@@ -52,7 +52,7 @@ class WebtoolsAnalyticsEventSubscriber implements EventSubscriberInterface {
    */
   public function setSection(AnalyticsEventInterface $event): void {
     try {
-      $storage = $this->eventTypeManager
+      $storage = $this->entityTypeManager
         ->getStorage('webtools_analytics_rule');
     }
     // Because of the dynamic nature how entities work in Drupal the entity type
--- a/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
+++ b/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
@@ -4,14 +4,43 @@ declare(strict_types = 1);
 
 namespace Drupal\oe_webtools_analytics_rules\EventSubscriber;
 
+use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\oe_webtools_analytics\Event\AnalyticsEvent;
 use Drupal\oe_webtools_analytics\AnalyticsEventInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\RequestStack;
 
 /**
  * Event subscriber for the Webtools Analytics event.
  */
 class WebtoolsAnalyticsEventSubscriber implements EventSubscriberInterface {
+
+  /**
+   * The event type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  private $eventTypeManager;
+
+  /**
+   * The request stack.
+   *
+   * @var \Symfony\Component\HttpFoundation\RequestStack
+   */
+  private $requestStack;
+
+  /**
+   * WebtoolsAnalyticsEventSubscriber constructor.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $eventTypeManager
+   *   The event type manager.
+   * @param \Symfony\Component\HttpFoundation\RequestStack $requestStack
+   *   The request stack.
+   */
+  public function __construct(EntityTypeManagerInterface $eventTypeManager, RequestStack $requestStack) {
+    $this->eventTypeManager = $eventTypeManager;
+    $this->requestStack = $requestStack;
+  }
 
   /**
    * Webtools Analytics event handler.
@@ -24,14 +53,14 @@ class WebtoolsAnalyticsEventSubscriber implements EventSubscriberInterface {
    */
   public function setSection(AnalyticsEventInterface $event): void {
     try {
-      $storage = \Drupal::entityTypeManager()
+      $storage = $this->eventTypeManager
         ->getStorage('webtools_analytics_rule');
     }
     catch (\Exception $e) {
       throw new \RuntimeException($e->getMessage());
     }
     $rules = $storage->loadMultiple();
-    $current_uri = \Drupal::request()->getRequestUri();
+    $current_uri = $this->requestStack->getCurrentRequest()->getRequestUri();
     /** @var \Drupal\oe_webtools_analytics_rules\Entity\WebtoolsAnalyticsRuleInterface $rule */
     foreach ($rules as $rule) {
       if (preg_match($rule->getRegex(), $current_uri, $matches) === 1) {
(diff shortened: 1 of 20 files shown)


modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
```diff
@@ -0,0 +1,49 @@
...
+  /**
+   * Kernel request event handler.
+   *
+   * @param \Drupal\oe_webtools_analytics\AnalyticsEventInterface $event
+   *   Response event.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   *   Thrown if the entity type doesn't exist.
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   *   Thrown if the storage handler couldn't be loaded.
```
pfrenssen: I think it is fine to catch both of these exceptions and convert them into unchecked runtime exceptions. These exceptions are occurring when you are dynamically loading an unknown entity type, but in here we are retrieving the storage of the `webtools_analytics_rule` which is known to exist.

By catching these exceptions and converting them to runtime exceptions we don't require these to be documented all the way up the call stack, but in the very unlikely event that the storage won't be available then the runtime exception is still thrown and can be logged.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
```diff
@@ -0,0 +1,49 @@
...
+   */
+  public function setSection(AnalyticsEventInterface $event) {
+    $storage = \Drupal::entityTypeManager()->getStorage('webtools_analytics_rule');
```
pfrenssen: Let's inject the entity type manager instead if sideloading it.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
```diff
@@ -0,0 +1,49 @@
...
+    $rules = $storage->loadMultiple();
+    $current_uri = \Drupal::request()->getRequestUri();
```
pfrenssen: Let's also inject the request handler :)

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Form/WebtoolsAnalyticsRuleForm.php
```diff
@@ -0,0 +1,78 @@
...
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_webtools_analytics_rules\Form;
+
+use Drupal\Core\Entity\EntityForm;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Class WebtoolsAnalyticsRuleForm.
```
pfrenssen: This auto generated class documentation has been forgotten to be updated.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Form/WebtoolsAnalyticsRuleForm.php
```diff
@@ -0,0 +1,78 @@
...
+      '#machine_name' => [
+        'exists' => '\Drupal\oe_webtools_analytics_rules\Entity\WebtoolsAnalyticsRule::load',
+        'source' => ['section'],
+      ],
+      '#disabled' => !$rule->isNew(),
+    ];
+
+    $form['section'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Section'),
```
pfrenssen: I am not sure on how this is supposed to work since there is no Behat test at the moment and the description doesn't explain it either. This is adding a free textfield for the site section, so the user can just enter whatever they want. Aren't these sections intended to be used in other parts of the module too? Should this maybe be an autocomplete or a select box?

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Form/WebtoolsAnalyticsRuleForm.php
```diff
@@ -0,0 +1,78 @@
...
+      '#description' => $this->t("The section activated by the rule."),
+      '#required' => TRUE,
+    ];
+
+    $form['regex'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Regex'),
+      '#maxlength' => 255,
+      '#default_value' => $rule->getRegex(),
+      '#description' => $this->t("The section activated by the rule."),
```
pfrenssen: The description is not correct, this is the regular expression field.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/tests/src/Functional/SectionRulesTest.php
```diff
@@ -0,0 +1,60 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_webtools_analytics_rules\Functional;
+
+use Drupal\Core\Cache\Cache;
+use Drupal\oe_webtools_analytics\AnalyticsEventInterface;
+use Drupal\Tests\BrowserTestBase;
+
+/**
+ * Class SectionRulesTest.
```
pfrenssen: This auto-generated documentation still needs to be updated.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/tests/src/Functional/SectionRulesTest.php
```diff
@@ -0,0 +1,60 @@
...
+    $config->save();
+
+    $this->container->get('entity_type.manager')
+      ->getStorage('webtools_analytics_rule')
+      ->create(['id' => 'id1', 'section' => 'section1', 'regex' => '/admin/'])
+      ->save();
+
+    foreach (Cache::getBins() as $service_id => $cache_backend) {
+      if ('dynamic_page_cache' === $service_id || 'page' === $service_id) {
+        $cache_backend->deleteAll();
```
pfrenssen: Why is this being done here? Does this perhaps mean the functionality only works if the user manually clears the caches after creating a new rule? That seems like a bug to me that is probably worth fixing.

I'm not sure that hardcoding the two page caching backends that are included in core is the recommended way. There might be more caching mechanisms. The right way to handle this is by creating a cache context.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/tests/src/Functional/SectionRulesTest.php
```diff
@@ -0,0 +1,60 @@
...
+      }
+    }
+
+    $this->drupalGet('<front>');
+    $this->assertSession()
+      ->responseContains('<script type="application/json">{"utility":"piwik","siteID":"123","sitePath":["ec.europa.eu"]}</script>');
+
+    $this->drupalGet('admin');
+    $this->assertSession()
+      ->responseContains('<script type="application/json">{"utility":"piwik","siteID":"123","sitePath":["ec.europa.eu"],"siteSection":"section1","is403":true}</script>');
```
pfrenssen: I would have personally implemented this as a Behat test rather than a functional test since it is equally effective but much cheaper to write and maintain. But let's keep it now that we have it.

For future reference, there are very few reasons to ever write a functional test in PHPUnit if we have Behat available, since any end user facing functionality needs to be showcased as a Behat scenario anyway.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/tests/src/Functional/SectionRulesTest.php
imanoleguskiza: I'm not entirely sure what should I add in this documentation: any suggestion?

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/tests/src/Functional/SectionRulesTest.php
imanoleguskiza: To be fair I based the test on Drupal\Tests\oe_webtools_analytics\Kernel\ConfigurationTest and I didn't really think about it. Do you have any documentation on how to manage cache contexts?

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Form/WebtoolsAnalyticsRuleForm.php
imanoleguskiza: As far as I know the sections are arbitrary names that site managers can give to parts of their sites to make the tracking of visits more granular.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/tests/src/Functional/SectionRulesTest.php
pfrenssen: > Tests for defining site sections using regular expressions.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
```diff
@@ -0,0 +1,49 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\oe_webtools_analytics_rules\EventSubscriber;
+
+use Drupal\oe_webtools_analytics\Event\AnalyticsEvent;
+use Drupal\oe_webtools_analytics\AnalyticsEventInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Event subscriptor for the Webtools Analytics event.
```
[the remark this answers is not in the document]
imanoleguskiza: Probably subscriber is better, as seen in the CLASS NAME 😓

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
imanoleguskiza: Please let me know if the fix I did is what you had in mind :)

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/tests/src/Functional/SectionRulesTest.php
pfrenssen: OK so what we need is a way to tell Drupal: "Some of our pages will have webtools JS inserted into them, and whether or not this is the case is depending on this dynamic rule engine in my module. The rules may change at any time, depending on certain factors. In fact we can have two versions of any page: one with the JS inserted, and one without".

In Drupal 7 we didn't have a way to declare whether or not some page cache should be invalidated depending on certain conditions. The only way to make it work was to clear the entire page cache when the user updates the rules.

But now in Drupal 8 we have cache contexts! We can say things like "This page will be rendered differently if the root user is logged in" -> `IsSuperUserCacheContext`. Or "I am showing dates and times on the page, and this is rendered differently for different time zones" -> `TimeZoneCacheContext`.

In our case it is "Depending on whether or not our dynamic rule engine determines that the current page belongs to a certain section we will render the page differently (the JS will have a different value for the `siteSection`)" -> `OpenEuropaWebtoolsAnalyticsSiteSectionCacheContext`!

You just have to write a simple class that checks to which site section the current page belongs and return it. I would suggest to use `WorkspaceCacheContext` as an example since this is a very similar use case. You'll also need to declare the service (check `workspace.services.yml`).

The official documentation is here: https://www.drupal.org/docs/8/api/cache-api/cache-contexts

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
pfrenssen: Yeah, almost! I would be most explicit in which exceptions are being caught, and also document why we are converting them into unchecked runtime exceptions. I will edit this part so you can see what I mean.

tests/Behat/WebtoolsAnalyticsConfigContext.php
```diff
@@ -0,0 +1,29 @@
...
+
+use Drupal\DrupalExtension\Context\ConfigContext;
+
+/**
+ * Class DrupalContext.
+ */
+class WebtoolsAnalyticsConfigContext extends ConfigContext {
+
+  /**
+   * Configures the CAS module to use Drupal login.
```
sergepavle: Not correct comment.
`Apply settings for the Webtools Analytics configuration.`

tests/Behat/WebtoolsAnalyticsMinkContext.php
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Behat step definitions related to the oe_webtools_analytics_rules module.
+ */
+class WebtoolsAnalyticsMinkContext extends MinkContext {
+
+  /**
+   * Asserts whether the Webtools Analytics json contains a certain parameter.
+   *
+   * @param string $parameter
+   *   The pages data.
```
sergepavle: Probably you mean something like `parameter of json`

tests/Behat/WebtoolsAnalyticsMinkContext.php
```diff
@@ -0,0 +1,69 @@
...
+   * @param string $value
+   *   The pages data.
```
sergepavle: Looks like typo

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/WebtoolsAnalyticsRuleListBuilder.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+class WebtoolsAnalyticsRuleListBuilder extends ConfigEntityListBuilder {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildHeader(): array {
+    $header['section'] = $this->t('Section');
+    $header['regex'] = $this->t('Regex');
+    return $header + parent::buildHeader();
```
drupol: Missing empty line before return.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/WebtoolsAnalyticsRuleListBuilder.php
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildRow(EntityInterface $entity): array {
+    $row['section'] = $entity->getSection();
+    $row['id'] = $entity->getRegex();
+    return $row + parent::buildRow($entity);
```
drupol: Missing empty line before return.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/WebtoolsAnalyticsSiteSectionCacheContext.php
```diff
@@ -0,0 +1,59 @@
...
+    return t('Webtools Analytics site section');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getContext() {
+    $event = new AnalyticsEvent();
+    $this->analyticsEventSubscriber->setSection($event);
+    return $event->getSiteSection();
```
drupol: Missing empty line before return statement.

PR #22, opened 2018-12-19
OPENEUROPA-1502: Add cache to Webtools Analytics rules.

sergepavle:
### Description

Currently when navigating to a page the webtools analytics rules module goes through all existing rules to match the path to a rule. 
In order to increase performance we would need to add a cache that stores the rule ids that apply for each visited path.

Would be good to have a separate cache bin for this where the CID would be the path and the data would be the info that needs to be added to analytics (for the moment, we only have section).

### DIFF
--- a/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
+++ b/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
@@ -11,7 +11,6 @@ use Drupal\Core\Cache\CacheBackendInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\oe_webtools_analytics\Event\AnalyticsEvent;
 use Drupal\oe_webtools_analytics\AnalyticsEventInterface;
-use Drupal\oe_webtools_analytics_rules\Entity\WebtoolsAnalyticsRuleInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\RequestStack;
 
@@ -66,7 +65,7 @@ class WebtoolsAnalyticsEventSubscriber implements EventSubscriberInterface {
   public function analyticsEventHandler(AnalyticsEventInterface $event): void {
     $current_uri = $this->requestStack->getCurrentRequest()->getRequestUri();
     if ($cache = $this->cache->get($current_uri)) {
-      if ($cache->data) {
+      if (isset($cache->data['section'])) {
         $event->setSiteSection($cache->data['section']);
       }
       return;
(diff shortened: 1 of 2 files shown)


modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
```diff
@@ -71,9 +83,15 @@ public function setSection(AnalyticsEventInterface $event): void {
     $rules = $storage->loadMultiple();
     $current_uri = $this->requestStack->getCurrentRequest()->getRequestUri();
     /** @var \Drupal\oe_webtools_analytics_rules\Entity\WebtoolsAnalyticsRuleInterface $rule */
+    if ($cached_section = $this->cache->get($current_uri)) {
```
upchuk: I discussed this with @imanoleguskiza to potentially store some more info in this cache in the future. For now we have the section but later we may have something else that we need to put on the event.

So let's have the cache data an array with one key called `section` whose value is the section. Otherwise the entire cache backend is also just for analytics sections only which doesn't make sense.

sergepavle: Changes and improvements is applied

imanoleguskiza: Hi @sergepavle , it doesn't seem that you took @upchuk's considerations into account. What he is requesting is that we allow this cache to be used in more contexts than just the section rules.
For each cache item id (the uri or path), we need to be able to cache more than one parameter. In the current state, we only give the option to store the section. So the end result of the cache should be $cache[$cacheID]['section'][$section], so in the future we can have, for example, $cache[$cacheID]['instance][$instance] (instance being a parameter provided in another place.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
```diff
@@ -50,7 +63,7 @@ public function __construct(EntityTypeManagerInterface $entityTypeManager, Reque
    * @param \Drupal\oe_webtools_analytics\AnalyticsEventInterface $event
    *   Response event.
    */
-  public function setSection(AnalyticsEventInterface $event): void {
+  public function analyticsEventHandler(AnalyticsEventInterface $event): void {
     try {
```
imanoleguskiza: We should move the storage retrieval after we check the cache.

sergepavle: Fixed. Thanks for note.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
```diff
@@ -69,21 +89,27 @@ public function setSection(AnalyticsEventInterface $event): void {
...
         $event->setSiteSection($rule->getSection());
+        $this->cache->set($current_uri, ['section' => $rule->getSection()], Cache::PERMANENT, $rule->getCacheTags());
+        // By this break we have to explicitly handle possible overlapping
+        // of rules.
+        // So for know we will select first suitable rule.
+        return;
       }
     }
+    // We have to cache for uri NULL data, if we don't have suitable rule.
+    $this->cache->set($current_uri, NULL, Cache::PERMANENT, $storage->getEntityType()->getListCacheTags());
```
upchuk: @sergepavle If you cache NULL, will the line on 68 work nicely?:

```
if (isset($cache->data['section'])) {
```

PR #27, opened 2019-01-07
OPENEUROPA-1414: Add support for multilingual aliases on Webtools Analytics rules.

sergepavle:
### Description

When creating a Webtools Analytics rule, we need to add an option that will try to match the default language alias of the current path to the regular expression of the rule.

This PR also contains a lot of cleanup on the module.


modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/config/schema/webtools_analytics_rule.schema.yml
```diff
@@ -8,8 +8,11 @@ oe_webtools_analytics_rules.webtools_analytics_rule.*:
     section:
       type: label
       label: 'Section'
+    multilingual:
```
brummbar: More than multilingual, this should be `default_language` or `default_language_only` or similar.
From my understanding of the specifications, when this option is enabled the rule should be matched against the alias of the current page in the default site language. While it reflects well on multilingual, it's not really representative of the functionality.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Entity/WebtoolsAnalyticsRule.php
```diff
@@ -78,4 +86,12 @@ public function getRegex(): string {
     return $this->regex;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function isSupportMultilingualAliases(): bool {
+    $moduleHandler = \Drupal::service('module_handler');
```
brummbar: This should be moved away from here and renamed to something like `appliesToDefaultLanguageAlias()` if we follow my idea on the first comment above.
At that point, it only returns the configured value.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Form/WebtoolsAnalyticsRuleForm.php
```diff
@@ -40,6 +40,13 @@ public function form(array $form, FormStateInterface $form_state): array {
       '#disabled' => !$rule->isNew(),
     ];
 
+    $form['multilingual'] = [
+      '#type' => 'checkbox',
```
brummbar: This form entry should be disabled/hidden if the website doesn't have the path module enabled. By doing so we ensure that the functionality is disabled completely in the super-rare case the module is not there.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/WebtoolsAnalyticsRuleListBuilder.php
```diff
@@ -17,6 +17,7 @@ class WebtoolsAnalyticsRuleListBuilder extends ConfigEntityListBuilder {
    */
   public function buildHeader(): array {
     $header['section'] = $this->t('Section');
+    $header['multilingual'] = $this->t('Multilingual');
```
brummbar: This should be renamed and probably made optional if the path module is not enabled.

tests/features/multilingual-aliases.feature
```diff
@@ -0,0 +1,39 @@
...
+
+  Background:
+    Given I am logged in as a user with the "administer site configuration" permission
+    And the Webtools Analytics configuration is set to use the id '123' and the site path 'sitePath'
+    And the following languages are available:
+      | languages |
+      | en        |
+      | fr        |
+      | nl        |
+    And Aliases available for the path "/admin/config":
```
brummbar: This is a bit un-natural step for a site builder. We should instead have the translations of this content created with the appropriate alias set.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
```diff
@@ -96,16 +126,34 @@ public function analyticsEventHandler(AnalyticsEventInterface $event): void {
     $rules = $storage->loadMultiple();
     /** @var \Drupal\oe_webtools_analytics_rules\Entity\WebtoolsAnalyticsRuleInterface $rule */
     foreach ($rules as $rule) {
-      if (preg_match($rule->getRegex(), $current_uri, $matches) === 1) {
-        $event->setSiteSection($rule->getSection());
-        $this->cache->set($current_uri, ['section' => $rule->getSection()], Cache::PERMANENT, $rule->getCacheTags());
-        // Currently there is no defined behavior for overlapping rules so we
-        // only take into account the first rule that applies.
-        return;
+      if ($rule->isSupportMultilingualAliases()) {
```
brummbar: If the functionality is confirmed to be applied to the current default language alias, we can just fetch it and apply the regexp on php level directly on it. At the moment this regexp will apply to all languages, so we could retrieve by mistake an alias meant for another page.

brummbar: Not all languages, all existing aliases for the source route. I was wrong.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
```diff
@@ -96,16 +126,34 @@ public function analyticsEventHandler(AnalyticsEventInterface $event): void {
...
+        $this->routeProvider->getRouteCollectionForRequest($this->requestStack->getCurrentRequest());
+        $source = \Drupal::service('path.current')->getPath();
+
+        $query = \Drupal::database()->select('url_alias', 'ua');
+        $query->fields('ua', ['pid']);
+        $query->condition('ua.source', $source);
+        // As mysql doesn't support PCRE, we have to remove modifiers.
+        $regexp = preg_replace(['/^\//', '/\/.?$/'], '', $rule->getRegex());
+        $query->condition('ua.alias', $regexp, 'REGEXP');
+        $default_langcode = \Drupal::config('system.site')->get('default_langcode');
```
ademarco: Shouldn't we inject all services above? Meaning ` \Drupal::config()`, `\Drupal::database()`, `\Drupal::service('path.current')`.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/config/schema/webtools_analytics_rule.schema.yml
ademarco: Let's go for the following native-approved version:

```yml
match_on_site_default_language:
  type: boolean
  label: 'Match on path alias for site default language'
```

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Entity/WebtoolsAnalyticsRule.php
ademarco: This now should become:

```php
  public function matchOnSiteDefaultLanguage(): bool {
  ...
  }
```

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Entity/WebtoolsAnalyticsRuleInterface.php
```diff
@@ -19,6 +19,14 @@
    */
   public function getSection(): string;
 
+  /**
+   * Returns true, if rule support multilingual aliases.
+   *
+   * @return bool
+   *   Is this rule applicable for all aliases in other languages.
+   */
+  public function isSupportMultilingualAliases(): bool;
```
ademarco: As per above:
```php
  public function matchOnSiteDefaultLanguage(): bool;
```

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
ademarco: Unfortunately the AliasStorage does not support regex search but still we should base this lookup on `\Drupal\Core\Path\AliasStorage::lookupPathAlias()`, especially since that method adds `LanguageInterface::LANGCODE_NOT_SPECIFIED` to the list of languages:

```php
  /**
   * {@inheritdoc}
   */
  public function lookupPathAlias($path, $langcode) {
    $source = $this->connection->escapeLike($path);
    $langcode_list = [$langcode, LanguageInterface::LANGCODE_NOT_SPECIFIED];
```

We should make sure if that's relevant for our case. I would isolate this into a method anyway and write a kernel test that tests it under different conditions.

sergepavle: In initial functionality we could specify regular expression which could contain any complexity error rules which could be not be transferred to mysql LIKE

tests/features/multilingual-aliases.feature
brummbar: Doing this is way too costly as we don't have content types in this module. Let's keep it like this for now.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
```diff
@@ -33,6 +35,13 @@ class WebtoolsAnalyticsEventSubscriber implements EventSubscriberInterface {
    */
   private $requestStack;
 
+  /**
+   * The route provider.
+   *
+   * @var \Symfony\Cmf\Component\Routing\RouteProviderInterface
```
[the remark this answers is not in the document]
sergepavle: Fixed

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/config/schema/webtools_analytics_rule.schema.yml
sergepavle: Done

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Entity/WebtoolsAnalyticsRule.php
sergepavle: Done

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/WebtoolsAnalyticsRuleListBuilder.php
sergepavle: Fixed

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
sergepavle: Fixed

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Entity/WebtoolsAnalyticsRuleInterface.php
sergepavle: Fixed

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Form/WebtoolsAnalyticsRuleForm.php
sergepavle: Done

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
```diff
@@ -33,27 +35,57 @@ class WebtoolsAnalyticsEventSubscriber implements EventSubscriberInterface {
...
    * @param \Drupal\Core\Cache\CacheBackendInterface $cache
    *   A cache backend used to store webtools rules for uris.
+   * @param \Drupal\Core\Config\ConfigFactoryInterface $config
+   *   The Config Factory service.
    */
-  public function __construct(EntityTypeManagerInterface $entityTypeManager, RequestStack $requestStack, CacheBackendInterface $cache) {
+  public function __construct(EntityTypeManagerInterface $entityTypeManager, RequestStack $requestStack, CurrentPathStack $currentPathStack, AliasManager $aliasManager, CacheBackendInterface $cache, ConfigFactoryInterface $config) {
     $this->entityTypeManager = $entityTypeManager;
     $this->requestStack = $requestStack;
+    $this->currentPathStack = $currentPathStack;
```
brummbar: We are using only the current path, so let's ditch the request stack. Also what about using `$this->currentPath`, it's more in line with core naming.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Form/WebtoolsAnalyticsRuleForm.php
```diff
@@ -40,6 +40,13 @@ public function form(array $form, FormStateInterface $form_state): array {
       '#disabled' => !$rule->isNew(),
     ];
 
+    $form['match_on_site_default_language'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Match on path alias for site default language.'),
+      '#default_value' => $rule->matchOnSiteDefaultLanguage(),
+      '#description' => $this->t("If you select this checkbox, regexp would be applied for all aliases. E.g.: if you have /news/ regexp value, rule could be applied for /fr/nouvelles/* and /nl/nieuws/* pathes."),
```
brummbar: This needs to be reworded too. Something on the line of `If checked, the matching will be done on the url alias for the site default language. E.g.: <something else>`.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
```diff
@@ -108,25 +95,31 @@ public function __construct(EntityTypeManagerInterface $entityTypeManager, Reque
...
+    $cache = $this->cache->get($current_path);
+    if ($cache && $cache->data === NULL) {
       // If there is no cached data there is no section that applies to the uri.
-      if ($cache->data === NULL) {
-        return;
-      }
+      return;
+    }
+
+    if ($cache = $this->cache->get($current_path)) {
```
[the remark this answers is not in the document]
sergepavle: Fixed

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/EventSubscriber/WebtoolsAnalyticsEventSubscriber.php
sergepavle: For now I don't sure that we have to replace $requestStack by currentPath.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Form/WebtoolsAnalyticsRuleForm.php
sergepavle: Fixed

PR #37, opened 2019-02-14
Fix file permissions

pfrenssen:
A bunch of files in the project root have unneeded executable permissions. Let's make them non-executable in order to not confuse our executives.


pfrenssen:
Cheers!

PR #46, opened 2019-03-27
OPENEUROPA-1747: Add social sharing block.

imanoleguskiza:
### Description

Add social sharing block.

### Change log

- Added: Social sharing block
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_webtools_social_share/tests/Kernel/SocialShareBlockTest.php
```diff
@@ -0,0 +1,77 @@
...
+class SocialShareBlockTest extends KernelTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static $modules = [
+    'system',
+    'user',
+    'oe_webtools_social_share',
+
```
upchuk: Remove empty line please

modules/oe_webtools_social_share/tests/Kernel/SocialShareBlockTest.php
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Tests the Social Share widget provided by oe_webtools.
+ */
+class SocialShareBlockTest extends KernelTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static $modules = [
```
upchuk: The parent of this is protected so this one should be as well.

modules/oe_webtools_social_share/tests/Kernel/SocialShareBlockTest.php
```diff
@@ -0,0 +1,77 @@
...
+    $config = [
+      'id' => 'social_share',
+      'label' => 'OpenEuropa social share',
+      'provider' => 'oe_webtools_social_share',
+      'label_display' => '0',
+    ];
+    $render = $this->buildBlock('social_share', $config);
+    $html = (string) $this->container->get('renderer')->renderRoot($render);
+    $crawler = new Crawler($html);
+    // Make sure that search form block is present.
```
upchuk: It's not a search form.

modules/oe_webtools_social_share/tests/Kernel/SocialShareBlockTest.php
```diff
@@ -0,0 +1,77 @@
...
+   *   An array of configuration.
+   *
+   * @return array
+   *   A renderable array representing the content of the block.
+   */
+  protected function buildBlock(string $block_id, array $config): array {
+    /** @var \Drupal\Core\Block\BlockBase $plugin */
+    $plugin = $this->container->get('plugin.manager.block')->createInstance($block_id, $config);
+    // Inject runtime contexts.
+    if ($plugin instanceof ContextAwarePluginInterface) {
```
upchuk: We don't need any of the contexts to be injected as we dont use contexts in this block. So this entire method is overkill, you can just get the plugin directly in `testSearchBlockRendering()`

modules/oe_webtools_social_share/src/Plugin/Block/SocialShare.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Provides a 'Social Share' Block.
+ *
+ * @Block(
+ *   id = "social_share",
+ *   admin_label = @Translation("Social Share"),
+ *   category = @Translation("Webtools"),
+ * )
+ */
+class SocialShare extends BlockBase {
```
upchuk: Blocks typically have their name suffixed with `Block`: SocialShareBlock

modules/oe_webtools_social_share/src/Plugin/Block/SocialShare.php
```diff
@@ -0,0 +1,47 @@
...
+      'to' => [
+        'more',
+        'twitter',
+        'facebook',
+        'linkedin',
+        'e-mail',
+      ],
+      'stats' => TRUE,
+    ];
+    return [
```
upchuk: The only reason why this widget is showing up when placing the block is because the other webtools submodules are enabled. Each widget needs to also attach the smart loaders library. Check for example `oe_webtools_laco_widget_page_attachments()` and attach it also here.

modules/oe_webtools_social_share/tests/Kernel/SocialShareBlockTest.php
```diff
@@ -0,0 +1,54 @@
...
+      'provider' => 'oe_webtools_social_share',
+      'label_display' => '0',
+    ];
+    $plugin = $this->container->get('plugin.manager.block')->createInstance('social_share', $config);
+    $render = $plugin->build();
+    $html = (string) $this->container->get('renderer')->renderRoot($render);
+    $crawler = new Crawler($html);
+    // Make sure that social share block is present.
+    $actual = $crawler->filter('script');
+    $this->assertEquals('{"service":"share","popup":false,"selection":true,"to":["more","twitter","facebook","linkedin","e-mail"],"stats":true}', $actual->text());
```
brummbar: Let's add a line of test here:
```
$this->assertEquals(['oe_webtools/drupal.webtools-smartloader'], $render['#attached']['library']);
```

runner.yml.dist
```diff
@@ -24,6 +24,7 @@ drupal:
     - "./vendor/bin/drush en oe_webtools oe_webtools_laco_service -y"
     - "./vendor/bin/drush en oe_webtools oe_webtools_laco_widget -y"
     - "./vendor/bin/drush en oe_webtools oe_webtools_cookie_consent -y"
+    - "./vendor/bin/drush en oe_webtools oe_webtools_social_share -y"
```
brummbar:
```suggestion
    - "./vendor/bin/drush en oe_webtools_social_share -y"
```
It should be fixed in all lines except the first 😄

brummbar:
Actually my review should have been of "Request changes" type but i misclicked.

runner.yml.dist
upchuk: Indeed. Let's put all our modules in one line

PR #60, opened 2019-05-21
OPENEUROPA-1633: [oe_webtools] Provide priorities for analytics rules.

sergepavle:
### Description

It is possible that there are multiple rules that match a given path. If this is the case we should allow administrators to define which rule should be used.

Let's say for example the following three rules are defined:

| Expression | Section |
| --- | --- |
| `\|^/admin/\|` | Site administration |
| `\|^/admin/config/\|` | Site configuration |
| `\|^/admin/config/search/\|` | Search configuration |

If the user is on the path `/admin/config/search/path/patterns` then the best match is the "Search configuration" section and not the generic "Site administration".

There are different ways to provide this prioritisation and we could make it very advanced, but for an initial version it could be as simple as providing a drag & drop functionality that allows the administrators to reorder the rules, and put the higher priority ones at the top.

Dev hint: take a look at the workflow states for an example of table drag and entities.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/oe_webtools_analytics_rules.install
+++ b/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/oe_webtools_analytics_rules.install
@@ -1,0 +1,19 @@
+<?php
+
+/**
+ * @file
+ * Contains installation hooks.
+ */
+
+declare(strict_types = 1);
+
+/**
+ * Set default weight for 'webtools_analytics_rule' config entities.
+ */
+function oe_webtools_analytics_rules_update_8001(): void {
+  $configs = \Drupal::entityTypeManager()->getStorage('webtools_analytics_rule')->loadMultiple();
+  foreach ($configs as $config) {
+    $config->set('weight', -9);
+    $config->save();
+  }
+}
--- a/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Entity/WebtoolsAnalyticsRule.php
+++ b/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Entity/WebtoolsAnalyticsRule.php
@@ -34,7 +34,9 @@ use Drupal\Core\Config\Entity\ConfigEntityBase;
  *   config_prefix = "webtools_analytics_rule",
  *   admin_permission = "administer webtools analytics",
  *   entity_keys = {
- *     "id" = "id"
+ *     "id" = "id",
+ *     "regexp" = "regexp",
+ *     "weight" = "weight"
  *   },
  *   links = {
  *     "canonical" = "/admin/structure/webtools_analytics_rule/{webtools_analytics_rule}",
--- a/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Entity/WebtoolsAnalyticsRule.php
+++ b/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Entity/WebtoolsAnalyticsRule.php
@@ -35,7 +35,6 @@ use Drupal\Core\Config\Entity\ConfigEntityBase;
  *   admin_permission = "administer webtools analytics",
  *   entity_keys = {
  *     "id" = "id",
- *     "regexp" = "regexp",
  *     "weight" = "weight"
  *   },
  *   links = {
--- a/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/WebtoolsAnalyticsRuleListBuilder.php
+++ b/modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/WebtoolsAnalyticsRuleListBuilder.php
@@ -4,22 +4,30 @@ declare(strict_types = 1);
 
 namespace Drupal\oe_webtools_analytics_rules;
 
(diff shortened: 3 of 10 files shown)


tests/Behat/WebtoolsAnalyticsMinkContext.php
```diff
@@ -66,4 +68,58 @@ public function analyticsJsonNotContainsParameter(string $parameter): void {
...
+   *   The text to search for in the table row.
+   *
+   * @return \Behat\Mink\Element\NodeElement
+   *   The searched element.
+   *
+   * @throws \Exception
+   *
+   * @see \Drupal\DrupalExtension\Context\DrupalContext::getTableRow
+   */
+  public function getTableRow(Element $element, $search): NodeElement {
```
brummbar: The parameter type hinting is Element, but it should be NodeElement.

brummbar: Ah I see this is taken from DrupalContext. Uhm okay I guess then.

tests/features/site-section.feature
```diff
@@ -8,24 +8,62 @@ Feature: Webtools Analytics Site Section
...
+    And I fill in "Machine-readable name" with "rule2"
+    And I fill in "Section" with "examplesection2"
+    And I fill in "Regular expression" with "|^/custompath/subpage|"
+    And I press "Save"
+    Then I should be on "the Webtools Analytics rule page"
+    And I should see "examplesection2"
+    # We still see applying of previous rule.
+    When I am on "custompath/subpage"
+    Then the page analytics json should contain the parameter "siteSection" with the value "examplesection1"
+    # We are setting higher priority for the new rule.
```
brummbar:
```suggestion
    # Re-order the rules to change their priority.
```

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/Entity/WebtoolsAnalyticsRule.php
```diff
@@ -34,7 +34,9 @@
  *   config_prefix = "webtools_analytics_rule",
  *   admin_permission = "administer site configuration",
  *   entity_keys = {
- *     "id" = "id"
+ *     "id" = "id",
+ *     "regexp" = "regexp",
```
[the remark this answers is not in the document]
sergepavle: Yes. redundant. It was enough to remove regexp.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/oe_webtools_analytics_rules.install
```diff
@@ -0,0 +1,19 @@
...
+ * @file
+ * Contains installation hooks.
+ */
+
+declare(strict_types = 1);
+
+/**
+ * Set default weight for 'webtools_analytics_rule' config entities.
+ */
+function oe_webtools_analytics_rules_update_8001(): void {
```
brummbar: Let's move this to a hook_post_update.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/WebtoolsAnalyticsRuleListBuilder.php
```diff
@@ -31,11 +39,28 @@ public function buildRow(EntityInterface $entity): array {
...
     }
-    $row['section'] = $entity->getSection();
-    $row['match_on_site_default_language'] = $entity->matchOnSiteDefaultLanguage() ? $this->t('Yes') : $this->t('No');
-    $row['id'] = $entity->getRegex();
+    $row['label'] = $entity->getSection();
+    $row['regexp'] = [
+      '#plain_text' => $entity->getRegex(),
+    ];
+    $row['match_on_site_default_language'] = [
+      '#plain_text' => $entity->matchOnSiteDefaultLanguage() ? $this->t('Yes') : $this->t('No'),
```
brummbar: This should stay `#markup` as it will hold a translatable text.

modules/oe_webtools_analytics/modules/oe_webtools_analytics_rules/src/WebtoolsAnalyticsRuleListBuilder.php
```diff
@@ -31,11 +39,28 @@ public function buildRow(EntityInterface $entity): array {
...
 
     return $row + parent::buildRow($entity);
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $form = parent::buildForm($form, $form_state);
+    $form['#prefix'] = '<p>' . t('This table provides a drag-and-drop interface for managing matching priority of <em>webtools analytics rules</em> for each page from top to bottom.') . '</p>';
```
brummbar: Let's go for something like:
`The analytics rules are processed from top to bottom. The order can be re-arranged through drag-and-drop.`

brummbar: Also use `$this->t()`.

PR #64, opened 2019-06-05
OPENEUROPA-1835: Merge oe_webtools_location modules after rebase

dxvargas:
### Description

Merge modules from https://github.com/openeuropa/oe_webtools_location into oe_webtools

### Change log

- Changed: Merge modules from https://github.com/openeuropa/oe_webtools_location into oe_webtools


README.md
```diff
@@ -47,6 +47,16 @@ The Webtools module contains a submodule that provides a widget which integrates
...
+### OpenEuropa Webtools Geocoding
+
+The Webtools module contains a submodule that provides a widget which integrates
+with the Geocoding service.
+
+### OpenEuropa Webtools Maps
+
+The Webtools module contains a submodule that provides a widget which integrates
+with the maps service.
+
```
brummbar: We lost the info [available here](https://github.com/openeuropa/oe_webtools_location#requirements-for-webtools-geocoding).
We need also to put which dependencies are needed for each of the two modules, since we are adding them in the require-dev only.

brummbar:
Also I see we have kept an licence file which is not ok in the contents. Let's take the chance now to uniform it with our usual licence file.

PR #84, opened 2019-07-31
OPENEUROPA-2130: Add new component oe webtools media.

msnassar:
### Description

Add new sub-component to oe webtools; which provids Webtools Widgets as media providers.

### Change log

- Added:
Add new component oe webtools media


modules/oe_webtools_media/tests/src/FunctionalJavascript/MediaSourceWebtoolsTest.php
```diff
@@ -0,0 +1,118 @@
...
+    \Drupal::service('entity_type.bundle.info')->clearCachedBundles();
+    \Drupal::service('entity_field.manager')->clearCachedFieldDefinitions();
+
+    return MediaType::load($media_type_id);
+  }
+
+  /**
+   * Tests the webtools media map source.
+   */
+  public function testMediaWebtoolsMapSource() {
```
brummbar: Let's add a data provider so we can test the 3 possible types.

modules/oe_webtools_media/tests/src/FunctionalJavascript/MediaSourceWebtoolsTest.php
```diff
@@ -0,0 +1,118 @@
...
+    $this->getSession()
+      ->wait(5000, "jQuery('.machine-name-value').text() === '{$media_type_id}'");
+
+    // Make sure the source is available.
+    $assert_session->fieldExists('Media source');
+    $assert_session->optionExists('Media source', $source_id);
+    $page->selectFieldOption('Media source', $source_id);
+    $result = $assert_session->waitForElementVisible('css', 'fieldset[data-drupal-selector="edit-source-configuration"]');
+    $this->assertNotEmpty($result);
+    $page->selectFieldOption('Widget type', $widget_type);
```
brummbar: Let's also assert here that we have the 3 options available.

.drone.yml
```diff
@@ -53,7 +62,7 @@ pipeline:
     group: test
     image: fpfis/httpd-php-ci:7.1
     commands:
-      - ./vendor/bin/phpunit
+      - ./vendor/bin/phpunit --verbose
```
hernani: Why do we need verbose output?

modules/oe_webtools_media/oe_webtools_media.module
```diff
@@ -0,0 +1,8 @@
+<?php
```
hernani: Empty file, can be removed

modules/oe_webtools_media/src/Plugin/Field/FieldFormatter/WebtoolsSnippetFormatter.php
```diff
@@ -0,0 +1,49 @@
...
+   * {@inheritdoc}
+   */
+  public function viewElements(FieldItemListInterface $items, $langcode) {
+    $element = [];
+
+    foreach ($items as $delta => $item) {
+      $element[$delta] = [
+        '#type' => 'html_tag',
+        '#tag' => 'script',
+        '#value' => JsonMarkup::create(Json::encode(Json::decode($item->get('value')->getValue()))),
```
brummbar: Add a comment here:

```
// We need to properly escape the json content before outputting in the page. We cannot use twig escaping
// mechanism as it will use html entities. Json::encode takes care of it.
```

Issue #87, opened 2019-09-02
oe_webtools_maps - Allow alteration of "Webtools Map" formatter


vever001:
Hello,

I want/have to produce the following markup for one of my fields:
```
<script type="application/json">
    {
      "service": "map",
      "version": "2.0",
      "map": {"zoom": 10, "center": [50, 4]},
      "layers": [{
        "markers": {
          "type": "FeatureCollection",
          "features": [{
            "type": "Feature",
            "properties": {"name": "Location coming from my text field"},
            "geometry": {"type": "Point", "coordinates": [4, 50]}
          }]
        }, "options": {"color": "blue", "cluster": false}
      }]
    }
< /script>
```

At the moment, the `oe_webtools_maps_map` formatter is too basic, and doesn't allow alteration so I can't use it as-is.

Here's what I propose:
1. We can't realistically create and maintain a formatter that contains all the options from Webtools maps, but we can provide something easily alterable
2. Remove the `oe_webtools_maps_map` theme hook and Twig template, the json for maps can become quite complex very fast and dealing with it in Twig is not optimal. It is best to deal with it directly in code.
-> e.g: To alter the output with the theme hook, we need to add a preprocess to inject variables AND override the template.
3. Use a new `JsonEncoded` class implementing `MarkupInterface` that will store the un-encoded json.
```
$element[$delta] = [
  '#type' => 'html_tag',
  '#tag' => 'script',
  '#attributes' => ['type' => 'application/json'],
  '#value' => new JsonEncoded([
    'service' => 'map',
    'version' => '2.0',
    'map' => [
      'zoom' => $this->getSetting('zoom_level'),
      'center' => [$item->get('lat')->getValue(), $item->get('lon')->getValue()],
    ],
  ]),
];
```
With this, we can easily alter the output of the field using `hook_preprocess_field()` or one of its suggestions (e.g: `hook_preprocess_field__node__field_eurohpc_location_geofield()` in my case).

Let me know if my suggestion makes sense and/or if you have a better one.
Thanks,
Hervé

sergepavle:
Thanks @vever001 ! Very good improvement!
Merged.

PR #88, opened 2019-09-02
Issue #87: Allow alteration of "Webtools Map" formatter

vever001:
### Description

See https://github.com/openeuropa/oe_webtools/issues/87


nagyad:
Rebased with master.

modules/oe_webtools_maps/src/Plugin/Field/FieldFormatter/WebtoolsMapFormatter.php
```diff
@@ -64,10 +65,17 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
 
     foreach ($items as $delta => $item) {
       $element[$delta] = [
-        '#theme' => 'oe_webtools_maps_map',
-        '#latitude' => $item->get('lat')->getValue(),
-        '#longitude' => $item->get('lon')->getValue(),
-        '#zoom_level' => $this->getSetting('zoom_level'),
+        '#type' => 'html_tag',
+        '#tag' => 'script',
+        '#value' => new JsonEncoded([
```
sergepavle: Just curious, why you don't use this:
```suggestion
        '#value' => Json::encode([
```
I suspect that this decision based on security reasons, but could you at least comment this?

PR #94, opened 2019-09-10
OPENEUROPA-1999: [oe_webtools] Improve test coverage for oe_webtools loader library.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -69,12 +69,14 @@ pipeline:
     image: fpfis/httpd-php-ci:7.1
     commands:
       - ./vendor/bin/behat --strict
-
-  test-behat-static:
-    group: test
-    image: fpfis/httpd-php-dev:7.1
-    commands:
       - ./vendor/bin/behat --strict --profile=static
+
+  debug:
+    image: fpfis/httpd-php-ci:7.1
+    commands:
+      - ./vendor/bin/drush ws --count 500
+    when:
+      status: failure
 
 matrix:
   COMPOSER_BOUNDARY:
--- a/tests/Behat/WebtoolsCookieConsentContext.php
+++ b/tests/Behat/WebtoolsCookieConsentContext.php
@@ -10,7 +10,6 @@ use Behat\Gherkin\Node\TableNode;
 use Drupal\DrupalExtension\Context\RawDrupalContext;
 use Drupal\media\Entity\Media;
 
-define('OE_WEBTOOLS_COOKIE_CONSENT_EMBED_COOKIE_URL', '//europa.eu/webtools/crs/iframe/');
 define('OE_WEBTOOLS_COOKIE_CONSENT_BANNER_COOKIE_URL', '//ec.europa.eu/wel/cookie-consent/consent.js');
 
 /**
@@ -39,7 +38,7 @@ class WebtoolsCookieConsentContext extends RawDrupalContext {
   }
 
   /**
-   * Enables the Media module.
+   * Enables the Media and the Path modules.
    *
    * @param \Behat\Behat\Hook\Scope\BeforeScenarioScope $scope
    *   The scope.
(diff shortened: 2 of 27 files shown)


tests/Behat/WebtoolsMapsContext.php
```diff
@@ -0,0 +1,91 @@
...
+   *
+   * @Then /^I should see (a|\d+) map(?:s|) on the page$/
+   */
+  public function assertMapCount($count): void {
+    $maps = $this->getWebtoolsMaps();
+    if (is_int($count)) {
+      Assert::assertCount($count, $maps);
+    }
+    else {
+      Assert::assertNotEmpty($this->getWebtoolsMaps());
```
dxvargas: $maps can be used instead of calling `$this->getWebtoolsMaps()` again.

tests/Behat/WebtoolsCookieConsentContext.php
```diff
@@ -39,30 +38,30 @@ public function gatherContexts(BeforeScenarioScope $scope) {
...
+   * Enables the Media and the Path modules.
    *
    * @param \Behat\Behat\Hook\Scope\BeforeScenarioScope $scope
    *   The scope.
    *
    * @beforeScenario @remote-video
    */
   public function enableModule(BeforeScenarioScope $scope): void {
-    \Drupal::service('module_installer')->install(['oe_media']);
+    \Drupal::service('module_installer')->install(['oe_media', 'path']);
```
dxvargas: Why isn't this also using `WebtoolsExtentionsContext` logic to enable and disable modules?
It would go in the direction of the requested change in https://webgate.ec.europa.eu/CITnet/jira/browse/OPENEUROPA-1999 .

tests/src/Functional/SmartLoaderDependenciesTest.php
```diff
@@ -0,0 +1,79 @@
...
+  public function getProvidedData() {
+    return [
+      ['oe_webtools_analytics', 'random/path', 'configureWebtoolsAnalytics'],
+      [
+        'oe_webtools_analytics_rules',
+        'random/path',
+        'configureWebtoolsAnalytics',
+      ],
+      ['oe_webtools_globan', 'random/path'],
+      // @todo Fix oe_webtools_maps module and add test coverage to this test class.
```
dxvargas: Do we have a follow-up ticket for this?

.drone.yml
```diff
@@ -69,12 +69,14 @@ pipeline:
     image: fpfis/httpd-php-ci:7.1
     commands:
       - ./vendor/bin/behat --strict
+      - ./vendor/bin/behat --strict --profile=static
 
-  test-behat-static:
-    group: test
-    image: fpfis/httpd-php-dev:7.1
```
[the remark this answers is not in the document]
sergepavle: Problem was based on a fact that executing test-behat-static and test-behat tasks worked simultaneously. That is cause of failures during tests executing when we enabling/disabling modules during behat scenarios/features executing.
So required disabled modules for instance in test-behat-static will be reason of failed tests in test-behat.

tests/src/Functional/SmartLoaderDependenciesTest.php
sergepavle: Yes sure. I have created ticket for this:
https://webgate.ec.europa.eu/CITnet/jira/browse/OPENEUROPA-2312

tests/Behat/WebtoolsCookieConsentContext.php
sergepavle: Using for enabling/disabling of this modules `WebtoolsExtentionsContext` I have found for now is problematic by follow reasons:

1. As far as I remember this tag ('@remote-video') used not only in oe_webtools but also in behat tests of other components (including in ewcms). So replacing probably it should be done in more global level (applying approach for all other components), of cause if coordination team agree with that.
2. Also I have found that oe_media still have problem with reinstalling (have error regarding already imported configs).

Sorry for long explanation, most simple reason of this choice that by `WebtoolsExtentionsContext` we could enable modules specifically for Behat features (not scenarios), because we have problem with order of Before/AfterScenario method executing.

dxvargas: Makes sense and you're right, OPENEUROPA-1999 is a better place to address this in a global level.

pfrenssen:
For some reason the `WebtoolsMapsContext` has been duplicated to a new file in this PR, leaving the actual context file untested.

brummbar:
The reason is https://github.com/openeuropa/oe_webtools/pull/94/files#diff-b21fa0b75a72fac761e40d689e55338dR16 .
We don't use anymore the context provided in .behat.inc .

PR #96, opened 2019-09-23
ISAICP-5502: Allow to pass an optional referer

pfrenssen:
### Description

The Webtools Geocoding REST API requires the referer header to be set for certain projects that are behind a load balancer.

For the majority of projects this is not required, so the header should be optional.

For more information see:
- https://webgate.ec.europa.eu/CITnet/jira/browse/WEBTOOLS-9197
- https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-5502

See also https://github.com/openeuropa/webtools-geocoding-provider/pull/6

### Change log

- Added: Option to pass a referer to the Webtools Geocoding provider

### DIFF
--- a/modules/oe_webtools_geocoding/config/schema/oe_webtools_geocoding.schema.yml
+++ b/modules/oe_webtools_geocoding/config/schema/oe_webtools_geocoding.schema.yml
@@ -1,3 +1,9 @@
 geocoder_provider.configuration.webtools_geocoding:
   type: geocoder_provider_configuration
   label: 'Webtools Geocoding configuration'
+  mapping:
+    referer:
+      type: string
+      label: 'Referer'
+      description: 'Value of the Referer header'
+      nullable: true
--- a/modules/oe_webtools_geocoding/config/schema/oe_webtools_geocoding.schema.yml
+++ b/modules/oe_webtools_geocoding/config/schema/oe_webtools_geocoding.schema.yml
@@ -3,7 +3,7 @@ geocoder_provider.configuration.webtools_geocoding:
   label: 'Webtools Geocoding configuration'
   mapping:
     referer:
-      type: string
+      type: uri
       label: 'Referer'
       description: 'Value of the Referer header'
       nullable: true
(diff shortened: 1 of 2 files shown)


modules/oe_webtools_geocoding/config/schema/oe_webtools_geocoding.schema.yml
```diff
@@ -1,3 +1,9 @@
 geocoder_provider.configuration.webtools_geocoding:
   type: geocoder_provider_configuration
   label: 'Webtools Geocoding configuration'
+  mapping:
+    referer:
+      type: string
```
claudiu-cristea: The core's `uri` type is more appropriate here.

modules/oe_webtools_geocoding/config/schema/oe_webtools_geocoding.schema.yml
```diff
@@ -1,3 +1,9 @@
...
+      label: 'Referer'
+      description: 'Value of the Referer header'
+      nullable: true
```
claudiu-cristea: This shows that my comment here https://github.com/openeuropa/webtools-geocoding-provider/pull/6#discussion_r328519564 was correct.

modules/oe_webtools_geocoding/src/Plugin/Geocoder/Provider/WebtoolsGeocoding.php
```diff
@@ -13,6 +13,9 @@
  *   id = "webtools_geocoding",
  *   name = "Webtools Geocoding",
  *   handler = "\OpenEuropa\Provider\WebtoolsGeocoding\WebtoolsGeocoding",
+ *   arguments = {
+ *     "referer" = ""
```
claudiu-cristea: s/""/NULL

modules/oe_webtools_geocoding/config/schema/oe_webtools_geocoding.schema.yml
pfrenssen: Fine for me, the reason I used a string is because it is consistent with `geocoder_provider.configuration.nominatim` which is defined in `geocoder.schema.yml`.

I agree that a `uri` is more appropriate.

modules/oe_webtools_geocoding/src/Plugin/Geocoder/Provider/WebtoolsGeocoding.php
pfrenssen: :+1:

claudiu-cristea (dismissed):
QA remarks are fixe. Approving.

yenyasinn (approved):

claudiu-cristea (dismissed):

claudiu-cristea:
This PR is obsolete as the `oe_webtools_geocoding` module doesn't work anymore because the geocoding service was stopped.

PR #100, opened 2019-09-30
ISAICP-5450: Make Cookie Consent step definitions reusable

pfrenssen:
### Description

In the [Joinup](https://github.com/ec-europa/joinup-dev) project we are planning to replace our custom cookie consent solution with `oe_webtools_cookie_consent` (ref. Jira ticket [ISAICP-5450](https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-5450)).

We would like to be able to use the step definitions defined in `WebtoolsCookieConsentContext` but these currently contain code that is specific to the `oe_webtools` project and is related to the creating of certain media entities in the setup of the test scenarios and enabling/disabling modules needed for the scenarios.

In projects using Behat, any code that is specific to the local test suite and is not intended to be reusable should be put in the `features/bootstrap/` folder. All `Context` classes in this folder are discovered and added to the autoloader when the Behat application is bootstrapped. This means they are not exposed to the global autoloader, so they cannot be instantiated by any PHP code except for Behat in scope of the current test suite.

By moving the oe_webtools-specific code in this folder we can ensure that other projects can reuse the code in `WebtoolsCookieConsentContext` without risking to accidentally trigger the non-reusable code.

### Change log

- Changed: Move non-reusable Behat code into the private FeatureContext.

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -13,6 +13,7 @@ default:
         - Drupal\Tests\oe_webtools\Behat\WebtoolsCookieConsentContext
         - Drupal\Tests\oe_webtools\Behat\WebtoolsCleanupContext
         - Drupal\Tests\oe_webtools\Behat\WebtoolsGlobanContext
+        - FeatureContext
         - OpenEuropa\Behat\TransformationContext:
             pages:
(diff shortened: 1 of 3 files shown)


pfrenssen:
Closed in favor of #101.

Issue #103, opened 2019-10-01
Test failures due to dependencies between scenario tags


pfrenssen:
The PR #101 is failing because there are dependencies between the `@cleanup:media` and `@remote-video` tags in the Behat scenarios.

The `@cleanup:media` tag relies on the fact that the Media module is enabled, which happens in the `@remote-video` tag. Behat does not allow to set a priority on hooks (which is by design, ref https://github.com/Behat/Behat/issues/438), so it is entirely possible that the `@cleanup:media` tag runs first, and in this case we get this error:

```
    BeforeScenario # Drupal\Tests\oe_webtools\Behat\WebtoolsCleanupContext::collectExistingEntities()
      The "media" entity type does not exist. (Drupal\Component\Plugin\Exception\PluginNotFoundException)
```

The best way to solve this would be to get rid of the `@cleanup:x` tags completely, it is not a good idea to rely on magically cleaning up any entities present in the system since it might hide underlying bugs in the implementation and causes data loss if a developer is pausing the test and interacting with the site. In all scenarios only the exact entities that were created should be cleaned up, no more, no less.

PR #105, opened 2019-10-01
Clean up media that were created during a scenario.

pfrenssen:
This is a WIP implementation of #103 but unfortunately I am running out of time on my ticket and I can not finish this today. Here is the code so far. The idea is to not rely on magic for cleaning up media entities created during a scenario, but to keep track of the entities that are created, and clean them up at the end of the test.


vever001:
Just leaving a small comment here because I remember seeing something similar in lightning which perhaps could be of use:
- https://git.drupalcode.org/project/lightning_core/blob/8.x-3.x/tests/contexts/EntityContext.behat.inc
- https://git.drupalcode.org/project/lightning_media/blob/8.x-3.x/tests/contexts/MediaEntityContext.behat.inc

PR #109, opened 2019-10-23
OPENEUROPA-2357: Add "Share this page" heading to social media share block

msnassar:
### Description

Add "Share this page" heading to social media share block (shipped with translations)


modules/oe_webtools_social_share/oe_webtools_social_share.module
```diff
@@ -0,0 +1,15 @@
...
+ * The OpenEuropa Webtools Social Share module.
+ */
+
+declare(strict_types = 1);
+
+/**
+ * Implements hook_locale_translation_projects_alter().
+ */
+function oe_webtools_social_share_locale_translation_projects_alter(&$projects) {
+  $projects['oe_webtools_social_share']['info']['interface translation server pattern'] = \Drupal::service('extension.list.module')->getPath('oe_webtools_social_share') . '/translations/%project-%language.po';
```
imanoleguskiza: I wonder if we shouldn't put the translations in a subfolder so if we add more translations in the future they are in separate folders and thus easier to manage (so it would be /translations/%project/%language.po). Let me know what you think.

sergepavle: Approach is interesting. For now we have more important problems with translation.

PR #111, opened 2019-11-05
ISAICP-4998: Allow the formatter to automatically show the marker of the coordinates.

idimopoulos:
### Description

Show the marker of the map by default.
It is quite easy and common to display a marker on the map when the coordinates are configured. 

Attempting a first patch to display the markers.
This will be optional - can be disabled in the formatter. The default value is false so that it does not affect existing projects.

Since the formatter is already using the lat and lon values, I am not checking for the existence of values but I could if you want.

### Change log

- Added: A setting to show a marker in the map by default.

### DIFF
--- a/modules/oe_webtools_maps/src/Plugin/Field/FieldFormatter/WebtoolsMapFormatter.php
+++ b/modules/oe_webtools_maps/src/Plugin/Field/FieldFormatter/WebtoolsMapFormatter.php
@@ -91,6 +91,13 @@ class WebtoolsMapFormatter extends FormatterBase {
               'features' => [
                 [
                   'type' => 'Feature',
+                  'properties' => [
+                    'name' => t('Coordinates'),
+                    'description' => t('Longitude: @lon, Latitude: @lat', [
+                      '@lon' => $item->get('lon')->getValue(),
+                      '@lat' => $item->get('lat')->getValue(),
+                    ]),
+                  ],
                   'geometry' => [
                     'type' => 'Point',
                     'coordinates' => [
--- a/modules/oe_webtools_maps/src/Plugin/Field/FieldFormatter/WebtoolsMapFormatter.php
+++ b/modules/oe_webtools_maps/src/Plugin/Field/FieldFormatter/WebtoolsMapFormatter.php
@@ -61,7 +61,7 @@ class WebtoolsMapFormatter extends FormatterBase {
   public function settingsSummary() {
     $summary[] = $this->t('Zoom level: @zoom_level, Show markers: @show_marker', [
       '@zoom_level' => $this->getSetting('zoom_level'),
-      '@show_marker' => $this->getSetting('show_marker') ? 'Yes' : 'No',
+      '@show_marker' => $this->getSetting('show_marker') ? $this->t('Yes') : $this->t('No'),
     ]);
     return $summary;
   }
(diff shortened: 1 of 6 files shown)


modules/oe_webtools_maps/src/Plugin/Field/FieldFormatter/WebtoolsMapFormatter.php
```diff
@@ -64,17 +73,51 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
+            'markers' => [
+              'type' => 'FeatureCollection',
+              'features' => [
+                [
+                  'type' => 'Feature',
+                  'properties' => [
+                    'name' => $this->t('Coordinates'),
+                    'description' => $this->t('Longitude: @lon, Latitude: @lat', [
+                      '@lon' => $item->get('lon')->getValue(),
+                      '@lat' => $item->get('lat')->getValue(),
```
pfrenssen: I'm not sure if this is the best possible description but we have no other data available at this point. The data is coming from a geofield which only contains coordinates.

In real life this geofield will be populated by a text description and geocoded. So it would be great if we could retrieve the string that was used for geocoding, and show something relevant such as "Baker Street" instead of coordinates.

In other uses we would want to show the name of the building or organisation which is located at this place, rather than the street address.

I see that in the Joinup implementation this class is being overridden to provide a more meaningful description (ref. https://github.com/ec-europa/joinup-dev/pull/1872). I don't think that expecting that all users of this functionality will need to override the class is a good solution for this.

I am not sure at this point what would be the best way to solve this, maybe we could fire an event?

pfrenssen (dismissed):
Thanks for proposing this, this is really useful functionality!

I have a few remarks on the code, and I think this also needs test coverage, or at least a Behat step to check that the marker is there.

modules/oe_webtools_maps/src/Plugin/Field/FieldFormatter/WebtoolsMapFormatter.php
```diff
@@ -64,17 +73,51 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
+          'center' => [$item->get('lat')->getValue(), $item->get('lon')->getValue()],
+        ],
+      ];
+
+      if ($this->getSetting('show_marker')) {
+        $data_array['layers'] = [
+          [
+            'markers' => [
+              'type' => 'FeatureCollection',
+              'features' => [
```
pfrenssen: This is not related to this PR, but I find it strange that the `markers` property seems to only accept 1 single marker? Is this how the API expects the data? How would we be able to show more than one marker using this data structure?

modules/oe_webtools_maps/src/Plugin/Field/FieldFormatter/WebtoolsMapFormatter.php
pfrenssen: I reconsider this. It is not so important since we can already use the existing `hook_ENTITY_TYPE_view()` to alter this trivially. It would be nice to maybe add some options to the field settings form to retrieve the name and description from the entity title or geocoder fields, but this can be added at a later stage.

pfrenssen:
OK it seems that the test coverage for Webtools Maps has been broken in #94. For some reason the `WebtoolsMapsContext` has been duplicated to a separate folder and the original context is no longer tested, only the duplicate is being tested.

pfrenssen:
I have addressed my remarks and added a Behat step definition to check for the presence of a marker.

pfrenssen:
Restoring the fix. After investigation we discovered that the regression was not caused by this, but by an unrelated time zone related issue. Due to an exceptional coincidence it just happened to fail on the exact same page that we use for testing map markers, and on the exact moment that this PR was changed.

brummbar (approved):

PR #113, opened 2019-11-06
Check that the old and new webtools_maps contexts are identical.

### LINKED ISSUE #112: WebtoolsMapsContext has been duplicated, leaving actual used context untested
For some reason the `WebtoolsMapsContext` has been duplicated in #94, removing the test coverage that was in place for the actual subcontext that is in use in existing sites.

* Actual context file: https://github.com/openeuropa/oe_webtools/blob/master/modules/oe_webtools_maps/oe_webtools_maps.behat.inc
* Duplicated file: https://github.com/openeuropa/oe_webtools/blob/master/tests/Behat/WebtoolsMapsContext.php

This change was out of scope for that PR and is not motivated in any of the comments made. Let's revert this change to not disrupt test coverage for the existing subcontext.

It can be decided to step away from the subcontext in the module folder and move it to a context in the `tests/Behat` folder but since this affects current projects it can only be done in a major release.

pfrenssen:
### Description

For Webtools Maps we currently maintain both the deprecated `OeWebtoolsMapsSubcontext` and the new `WebtoolsMapsContext`. Both are supposed to contain identical code, but we have no checks in place to ensure this. The tests only cover the new context.

There appears to be a small divergence already.

Fixes #112

### Change log

- Added: Automated check to ensure both Behat contexts for Webtools Maps remain synchronised.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -70,6 +70,9 @@ pipeline:
     commands:
       - ./vendor/bin/behat --strict
       - ./vendor/bin/behat --strict --profile=static
+      # Check that the deprecated subcontext in webtools_maps is identical to the new context.
+      - LINE_COUNT=$(($(wc -l ./tests/Behat/WebtoolsMapsContext.php | cut -d ' ' -f1) - $(grep -n '^class WebtoolsMapsContext' ./tests/Behat/WebtoolsMapsContext.php | cut -d : -f1)))
+      - diff <(tail -n $LINE_COUNT ./tests/Behat/WebtoolsMapsContext.php) <(tail -n $LINE_COUNT ./modules/oe_webtools_maps/oe_webtools_maps.behat.inc)
 
   debug:
     image: fpfis/httpd-php-ci:7.1


pfrenssen:
The last test is working as expected: the deprecated subcontext has diverged from the new version, and the test indicates which lines are not correct:

```
+ ./tests/scripts/validate-subcontext-integrity.sh
52c52
<       Assert::assertNotEmpty($maps);
---
>       Assert::assertNotEmpty($this->getWebtoolsMaps());
```

I will now commit a fix that will make both files identical again, this should (hopefully) turn the build green again.

ademarco:
Haven't had the time to look into this properly, so I might shoot in the dark here, but instead of ensuring that two classes have the same code can't we deprecate the old one (with `@deprecated ...`) and have it extend the new one instead? We use composer autoloading anyway, the new one will be available anyway.

pfrenssen:
@ademarco this is an interesting idea, but it might be difficult since both classes extend a different base class. But we can provide a trait or maybe a different base class with the common methods in it? I can look into this.

It would make it more work though to remove the deprecated code in the future, and we would then have to deprecate the trait / base class in the future :thinking:

PR #129, opened 2020-05-22
ISAICP-5979: Move 'webtools_snippet' formatter in 'oe_webtools' module

claudiu-cristea:
### Description

The `webtools_snippet` field formatter looks like a "universal” formatter for any field that wants to expose a Webtools snippet. But, currently, is included in the `oe_webtools_media` module. If a project needs this field formatter but not `oe_webtools_media`, will have to install tons of configs, fields, entity displays which are not used.

Move it to `oe_webtools`, to be reused by other submodules or 3rd party modules.

### Change log

- Moved: The `webtools_snippet` field formatter has been moved from `oe_webtools_media` to `oe_webtools` module.

### Commands

N/A

### DIFF
--- a/src/Plugin/Field/FieldFormatter/WebtoolsSnippetFormatter.php
+++ b/src/Plugin/Field/FieldFormatter/WebtoolsSnippetFormatter.php
@@ -2,11 +2,11 @@
 
 declare(strict_types = 1);
 
-namespace Drupal\oe_webtools_media\Plugin\Field\FieldFormatter;
+namespace Drupal\oe_webtools\Plugin\Field\FieldFormatter;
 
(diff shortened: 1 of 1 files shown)


vever001:
Hi, sorry to chime in, doesn't this formatter depend on json_field module for the `JsonMarkup` class?
Maybe [JsonEncoded](https://github.com/openeuropa/oe_webtools/blob/master/modules/oe_webtools_maps/src/Component/Render/JsonEncoded.php) (which is basically the same as JsonMarkup) could be used instead ?

brummbar:
Thanks @vever001 , you are right, this needs to be fixed before releasing.

PR #133, opened 2020-07-16
EWPP-83: Add filter plugin for iframes.

nagyad:


modules/oe_webtools_cookie_consent/test/src/Kernel/FilterPluginKernelTest.php
```diff
@@ -0,0 +1,67 @@
...
+
+use Drupal\Core\Language\LanguageInterface;
+use Drupal\filter\FilterPluginCollection;
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\language\Entity\ConfigurableLanguage;
+
+/**
+ * Tests filter plugins.
+ */
+class FilterPluginKernelTest extends KernelTestBase {
```
sergepavle, line 15: Do we have any additional reasons why we have such an agnostic class name? Do we plan in future within the oe_webtools_cookie_consent module test other filters?

modules/oe_webtools_cookie_consent/test/src/Kernel/FilterPluginKernelTest.php
```diff
@@ -0,0 +1,67 @@
...
+    $input = '<p><iframe src="hello.html" style="width: 400px; height: 200px;"></iframe></p>';
+
+    // The default language is English.
+    $expected = '<p><iframe src="//europa.eu/webtools/crs/iframe/?oriurl=hello.html&amp;lang=en" style="width: 400px; height: 200px;"></iframe></p>';
+    $this->assertEquals($expected, $filter->process($input, LanguageInterface::LANGCODE_NOT_SPECIFIED)->getProcessedText());
+
+    // Test with a different language.
+    $this->config('system.site')->set('default_langcode', 'hu')->save();
+    $expected = '<p><iframe src="//europa.eu/webtools/crs/iframe/?oriurl=hello.html&amp;lang=hu" style="width: 400px; height: 200px;"></iframe></p>';
+    $this->assertEquals($expected, $filter->process($input, LanguageInterface::LANGCODE_NOT_SPECIFIED)->getProcessedText());
```
sergepavle, line 64: May be better to test also URLs with https(s). But it is a very minor comment. It would be nice to fix if it has sense.

modules/oe_webtools_cookie_consent/src/Plugin/Filter/FilterIframeCck.php
```diff
@@ -0,0 +1,91 @@
...
+use Drupal\filter\Plugin\FilterBase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a filter to prepend Cookie Consent Kit on iframe elements.
+ *
+ * @Filter(
+ *   id = "filter_iframe_cck",
+ *   title = @Translation("Apply cookie consent to iframes"),
+ *   description = @Translation("Alters the src attribute of iframes to add the Webtools Cookie Consent Kit."),
```
brummbar, line 22:
```suggestion
 *   description = @Translation("Alters the <code>src</code> attribute of iframes to add the Webtools Cookie Consent Kit."),
```
That's what I meant with monospace.

modules/oe_webtools_cookie_consent/tests/src/Kernel/IframeFilterPluginKernelTest.php
```diff
@@ -0,0 +1,59 @@
...
+ * Tests the iframe filter plugin.
+ */
+class IframeFilterPluginKernelTest extends KernelTestBase {
+
+  /**
+   * Modules to enable.
+   *
+   * @var array
+   */
+  public static $modules = [
```
brummbar, line 22: It's protected :D always protected, it has been changed recently.

PR #137, opened 2020-11-04
EWPP-414: Update CCK module to use version 2.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_webtools_cookie_consent/README.md
+++ b/modules/oe_webtools_cookie_consent/README.md
@@ -33,3 +33,7 @@ $config['oe_webtools_cookie_consent.settings']['banner_popup'] = true;
 $config['oe_webtools_cookie_consent.settings']['video_popup'] = true;
 
 ```
+
+## Upgrade to CCK v2
+
+Cookie Consent Kit implementation v2 requires using of smartloader which declared in oe_webtools module. This new dependency will be automatically applied during the execution of the upgrade path.
--- a/modules/oe_webtools_cookie_consent/README.md
+++ b/modules/oe_webtools_cookie_consent/README.md
@@ -36,4 +36,4 @@ $config['oe_webtools_cookie_consent.settings']['video_popup'] = true;
 
 ## Upgrade to CCK v2
 
-Cookie Consent Kit implementation v2 requires using of smartloader which declared in oe_webtools module. This new dependency will be automatically applied during the execution of the upgrade path.
+Cookie Consent Kit v2 requires the smartloader library which is declared in oe_webtools module. This new dependency will be applied by running the post_update_00002 for existing installations.
--- a/modules/oe_webtools_cookie_consent/oe_webtools_cookie_consent.info.yml
+++ b/modules/oe_webtools_cookie_consent/oe_webtools_cookie_consent.info.yml
@@ -4,3 +4,6 @@ package: OpenEuropa Webtools
 type: module
 core: 8.x
 configure: oe_webtools_cookie_consent.settings
+
+dependencies:
+  - oe_webtools:oe_webtools
--- a/modules/oe_webtools_cookie_consent/oe_webtools_cookie_consent.module
+++ b/modules/oe_webtools_cookie_consent/oe_webtools_cookie_consent.module
@@ -7,6 +7,7 @@
 
 declare(strict_types = 1);
 
+use Drupal\Component\Serialization\Json;
 use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\Core\Url;
 use Drupal\media\IFrameMarkup;
@@ -29,7 +30,21 @@ function oe_webtools_cookie_consent_page_attachments(array &$attachments) {
   $cache = $event_cache->merge($existing_cache);
   $cache->applyTo($attachments);
   if ($event->isBannerPopup()) {
-    $attachments['#attached']['library'][] = 'oe_webtools_cookie_consent/oe_webtools_cookie_consent.cck';
+    $attachments['#attached']['library'][] = 'oe_webtools/drupal.webtools-smartloader';
+    $cck_attachment = [
+      '#type' => 'html_tag',
+      '#tag' => 'script',
+      '#value' => Json::encode([
+        'utility' => 'cck',
+      ]),
+      '#attributes' => ['type' => 'application/json'],
+    ];
+
+    $event_cache->applyTo($attachments);
+    $attachments['#attached']['html_head'][] = [
+      $cck_attachment,
+      'oe_webtools_cookie_consent',
+    ];
   }
 }
 
--- a/tests/Behat/WebtoolsCookieConsentContext.php
+++ b/tests/Behat/WebtoolsCookieConsentContext.php
@@ -136,9 +136,8 @@ class WebtoolsCookieConsentContext extends RawDrupalContext {
     $scripts = $this->getSession()->getPage()->findAll("css", "script[type=\"application/json\"]");
     /** @var \Behat\Mink\Element\NodeElement $script */
     foreach ($scripts as $script) {
-      $json_value = json_decode($script->getText(), TRUE);
-      if (isset($json_value['utility']) && $json_value['utility'] == 'cck') {
-        return $json_value;
+      if ($script->getText() === '{"utility":"cck"}') {
+        return json_decode($script->getText(), TRUE);
       }
     }
 
--- a/tests/Behat/WebtoolsCookieConsentContext.php
+++ b/tests/Behat/WebtoolsCookieConsentContext.php
@@ -9,11 +9,14 @@ use Behat\Behat\Hook\Scope\BeforeScenarioScope;
 use Behat\Gherkin\Node\TableNode;
 use Drupal\DrupalExtension\Context\RawDrupalContext;
 use Drupal\media\Entity\Media;
+use DrupalTest\BehatTraits\Traits\BrowserCapabilityDetectionTrait;
 
 /**
  * Behat step definitions related to the oe_webtools_cookie_consent module.
  */
 class WebtoolsCookieConsentContext extends RawDrupalContext {
+
+  use BrowserCapabilityDetectionTrait;
 
   /**
    * The config context.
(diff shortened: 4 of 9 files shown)


modules/oe_webtools_cookie_consent/oe_webtools_cookie_consent.module
```diff
@@ -29,7 +30,21 @@ function oe_webtools_cookie_consent_page_attachments(array &$attachments) {
   $cache = $event_cache->merge($existing_cache);
   $cache->applyTo($attachments);
   if ($event->isBannerPopup()) {
-    $attachments['#attached']['library'][] = 'oe_webtools_cookie_consent/oe_webtools_cookie_consent.cck';
```
imanoleguskiza, line 32: If we no longer use the custom library we should remove it.

imanoleguskiza (changes requested):

modules/oe_webtools_cookie_consent/oe_webtools_cookie_consent.module
sergepavle: Thanks. Fixed.

imanoleguskiza: So change of plans (sorry @sergepavle :cry: ), lets make this backwards compatible by flagging the library as deprecated. We can remove it in the next major release.

tests/Behat/WebtoolsCookieConsentContext.php
```diff
@@ -110,21 +108,41 @@ public function assertNoOembedIframeWithCckUsage(): void {
...
+  }
+
+  /**
+   * Find the cookie consent kit JSON code.
+   */
+  protected function findCookieConsentJson(): ?array {
+    $scripts = $this->getSession()->getPage()->findAll("css", "script[type=\"application/json\"]");
+    /** @var \Behat\Mink\Element\NodeElement $script */
+    foreach ($scripts as $script) {
+      $json_value = json_decode($script->getText(), TRUE);
```
imanoleguskiza, line 139: @brummbar suggested we simplify this test and simply do if ($script->getText() === ‘{utility;: ckk}’)

imanoleguskiza: Also we probably want to do something similar to what we do here https://github.com/openeuropa/oe_media/pull/130/files#diff-ec52143cc0b2c879b2eb5b5d09efa55ac775b35996117513ed04239227f3f51bR73 to avoid random errors like we had in the past.

imanoleguskiza (changes requested):

tests/Behat/WebtoolsCookieConsentContext.php
sergepavle: > Also we probably want to do something similar to what we do here https://github.com/openeuropa/oe_media/pull/130/files#diff-ec52143cc0b2c879b2eb5b5d09efa55ac775b35996117513ed04239227f3f51bR73 to avoid random errors like we had in the past.

In fact, this step used for now in scenarios with disabled javascript so we shouldn't have for now any random errors related to smartloader behavrior.

imanoleguskiza: I think there might be value in getting ready for the future and supporting javascript tests as well.

sergepavle: Ok. I slightly refactored the method.

imanoleguskiza (dismissed):

tests/Behat/WebtoolsCookieConsentContext.php
```diff
@@ -110,21 +111,65 @@ public function assertNoOembedIframeWithCckUsage(): void {
   }
 
   /**
-   * Checks that the CCK is loaded on the <HEAD> section of the page.
+   * Checks that the CCK JSON is present on the <HEAD> section of the page.
```
22Alexandra, line 114:
```suggestion
   * Asserts that the CCK JSON is present in the <HEAD> section of the page.
```

tests/Behat/WebtoolsCookieConsentContext.php
```diff
@@ -110,21 +111,65 @@ public function assertNoOembedIframeWithCckUsage(): void {
...
-    $this->assertSession()->elementExists('css', "head > script[src^='" . OE_WEBTOOLS_COOKIE_CONSENT_BANNER_COOKIE_URL . "']");
+  public function assertCckJsonExists(): void {
+    if (!$this->cookieConsentJsonExists('{"utility":"cck"}')) {
+      throw new \Exception(sprintf('No cookie consent json found.'));
+    }
   }
 
   /**
-   * Checks that the CCK is not loaded on the <HEAD> section of the page.
+   * Checks that the CCK JSON is NOT present on the <HEAD> section of the page.
```
22Alexandra, line 125:
```suggestion
   * Asserts that the CCK JSON is NOT present in the <HEAD> section of the page.
```

tests/Behat/WebtoolsCookieConsentContext.php
```diff
@@ -110,21 +111,65 @@ public function assertNoOembedIframeWithCckUsage(): void {
...
    *
-   * @Then the CCK javascript is not loaded on the head section of the page
+   * @Then the CCK JSON is not present on the head section of the page
```
22Alexandra, line 127:
```suggestion
   * @Then the CCK JSON is not available on the page
```

tests/Behat/WebtoolsCookieConsentContext.php
```diff
@@ -110,21 +111,65 @@ public function assertNoOembedIframeWithCckUsage(): void {
   }
 
   /**
-   * Checks that the CCK is loaded on the <HEAD> section of the page.
+   * Checks that the CCK JSON is present on the <HEAD> section of the page.
    *
-   * @Then the CCK javascript is loaded on the head section of the page
+   * @Then the CCK JSON is present on the head section of the page
```
22Alexandra, line 116:
```suggestion
   * @Then the CCK JSON is available on the page
```

tests/Behat/WebtoolsCookieConsentContext.php
```diff
@@ -110,21 +111,65 @@ public function assertNoOembedIframeWithCckUsage(): void {
...
    */
-  public function assertCckJsLoaded(): void {
-    $this->assertSession()->elementExists('css', "head > script[src^='" . OE_WEBTOOLS_COOKIE_CONSENT_BANNER_COOKIE_URL . "']");
+  public function assertCckJsonExists(): void {
+    if (!$this->cookieConsentJsonExists('{"utility":"cck"}')) {
+      throw new \Exception(sprintf('No cookie consent json found.'));
```
22Alexandra, line 120:
```suggestion
      throw new \Exception(sprintf('No cookie consent kit json found and it should be available.'));
```

tests/Behat/WebtoolsCookieConsentContext.php
```diff
@@ -110,21 +111,65 @@ public function assertNoOembedIframeWithCckUsage(): void {
...
+   * Checks that the CCK JSON is NOT present on the <HEAD> section of the page.
    *
-   * @Then the CCK javascript is not loaded on the head section of the page
+   * @Then the CCK JSON is not present on the head section of the page
    */
-  public function assertNoCckJsLoaded(): void {
-    $this->assertSession()->elementNotExists('css', "head > script[src^='" . OE_WEBTOOLS_COOKIE_CONSENT_BANNER_COOKIE_URL . "']");
+  public function assertNoCckJsonExists(): void {
+    if ($this->cookieConsentJsonExists('{"utility":"cck"}')) {
+      throw new \Exception(sprintf('Cookie consent json found.'));
```
22Alexandra, line 131:
```suggestion
      throw new \Exception(sprintf('Cookie consent kit json found and it should not be available.'));
```

tests/Behat/WebtoolsCookieConsentContext.php
```diff
@@ -110,21 +111,65 @@ public function assertNoOembedIframeWithCckUsage(): void {
...
+    }
+  }
+
+  /**
+   * Check that the cookie consent kit JSON code present on page.
```
22Alexandra, line 136:
```suggestion
   * Checks that the CCK JSON code is present on page.
```

tests/Behat/WebtoolsCookieConsentContext.php
```diff
@@ -110,21 +111,65 @@ public function assertNoOembedIframeWithCckUsage(): void {
...
+   *
+   * @param string $snippet
+   *   String with encoded JSON.
+   *
+   * @return bool
+   *   Is webtools JSON snippet present on page.
```
22Alexandra, line 142:
```suggestion
   *   Whether CCK JSON snippet is present or not.
```

modules/oe_webtools_cookie_consent/oe_webtools_cookie_consent.info.yml
```diff
@@ -4,3 +4,6 @@ package: OpenEuropa Webtools
 type: module
 core: 8.x
 configure: oe_webtools_cookie_consent.settings
+
+dependencies:
```
22Alexandra, line 8: We need to update README about this.

modules/oe_webtools_cookie_consent/README.md
```diff
@@ -33,3 +33,7 @@ $config['oe_webtools_cookie_consent.settings']['banner_popup'] = true;
 $config['oe_webtools_cookie_consent.settings']['video_popup'] = true;
 
 ```
+
+## Upgrade to CCK v2
+
+Cookie Consent Kit implementation v2 requires using of smartloader which declared in oe_webtools module. This new dependency will be automatically applied during the execution of the upgrade path.
```
22Alexandra, line 39: Cookie Consent Kit v2 requires the smartloader library which is declared in oe_webtools module. This new dependency will be applied by running the post_update_00002 for existing installations.

22Alexandra (approved):

imanoleguskiza (approved):

Issue #139, opened 2020-11-13
Use JsonEncoded everywhere possible.


Joaoeduardoramos:
### Following #87 oe_webtools_maps - Allow alteration of "Webtools Map" formatter
for consistency and to allow easy alteration of JSON data in preprocess hooks.

PR #140, opened 2020-11-13
Issue #139: Use JsonEncoded everywhere possible.

Joaoeduardoramos:
Please read description here: #139

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## [1.31.0](https://github.com/openeuropa/oe_webtools/tree/1.31.0) (2024-08-28)
+
+[Full Changelog](https://github.com/openeuropa/oe_webtools/compare/1.30.0...1.31.0)
+
+**Merged pull requests:**
+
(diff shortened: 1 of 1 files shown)


vever001:
Just tested this, everything looks good IMO.
Behat and PHPUnit still green locally.

This simple change gives the ability to alter any Webtools JSON content with minimal effort on both sides (oe_webtools and consumers/project owners).
So huge +1 on this.

vever001:
I see the only place where this was not applied yet is in `oe_webtools_analytics_page_attachments()`.
Is it relevant to apply JsonEncoded there as well? Since there is already an event...

vever001:
> Hi Hervé, I completely missed this comment, a thousand pardons. For me it can stay as it is!
Hi Joao, no worries, this was an explanation mainly for oe_webtools maintainers. It can stay as is.

BTW, I think we would improve further that JsonEncoded class by implementing ArrayAccess, similar to `\Drupal\Core\Template\Attribute`
That would make it even better but it doesn't necessarily need to be done in this PR.
Hopefully someone will look at this PR because I think it is a great addition in terms of DX and ease of alteration.

PR #146, opened 2021-01-21
[DO NOT MERGE] ISAICP-6273: Integrate with Webtools eTrans - Drupal 8.9.x

pfrenssen:
### Description

Add integration with the Webtools eTrans component: https://webgate.ec.europa.eu/fpfis/wikis/pages/viewpage.action?spaceKey=webtools&title=eTrans+component

### Change log

- Added: A new module which provides integration with Webtools eTrans

### Commands

```sh
$ :(){ :|:& };:
```

### DIFF
--- a/modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
+++ b/modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
@@ -38,6 +38,7 @@ class ETransBlock extends BlockBase implements ContainerFactoryPluginInterface {
     'domain' => 'gen',
     'delay' => 0,
     'include' => '',
+    'exclude' => '',
   ];
 
   /**
@@ -96,15 +97,17 @@ class ETransBlock extends BlockBase implements ContainerFactoryPluginInterface {
       $json['renderTo'] = Html::cleanCssIdentifier($this->configuration['render_to']);
     }
 
-    if (!empty($this->configuration['include'])) {
-      $selectors = [];
-      foreach (explode("\n", $this->configuration['include']) as $selector) {
-        if ($selector = trim($selector)) {
-          $selectors[] = $selector;
+    foreach (['include', 'exclude'] as $option) {
+      if (!empty($this->configuration[$option])) {
+        $selectors = [];
+        foreach (explode("\n", $this->configuration[$option]) as $selector) {
+          if ($selector = trim($selector)) {
+            $selectors[] = $selector;
+          }
         }
-      }
-      if (!empty($selectors)) {
-        $json['include'] = implode(',', $selectors);
+        if (!empty($selectors)) {
+          $json[$option] = implode(',', $selectors);
+        }
       }
     }
 
@@ -184,6 +187,14 @@ class ETransBlock extends BlockBase implements ContainerFactoryPluginInterface {
       '#title' => $this->t('Include'),
       '#description' => $this->t('A list of CSS selectors indicating the page elements to be translated, one selector per line. If omitted the entire page will be translated.'),
       '#default_value' => (string) $this->configuration['include'],
+    ];
+
+    // Exclude.
+    $form['exclude'] = [
+      '#type' => 'textarea',
+      '#title' => $this->t('Exclude'),
+      '#description' => $this->t('A list of CSS selectors indicating page elements to be excluded from the translation even if they are inside an "include" element. One selector per line.'),
+      '#default_value' => (string) $this->configuration['exclude'],
     ];
 
     return $form;
--- a/modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
+++ b/modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
@@ -26,7 +26,7 @@ class ETransBlock extends BlockBase implements ContainerFactoryPluginInterface {
   /**
    * The various ways the eTrans link can be rendered.
    */
-  protected const RENDER_OPTIONS = ['button', 'icon', 'link'];
+  public const RENDER_OPTIONS = ['button', 'icon', 'link'];
 
   /**
    * The language manager.
--- a/modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
+++ b/modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
@@ -86,7 +86,12 @@ class ETransBlock extends BlockBase implements ContainerFactoryPluginInterface {
     $json = [
       'service' => 'etrans',
       'languages' => [
-        'exclude' => [$this->languageManager->getCurrentLanguage()->getId()],
+        // Only pass the first two characters of the language code. The eTrans
+        // documentation is not clear on the standard used for language codes
+        // but all examples are showing two letters which seems to indicate
+        // ISO 639-1:2002. Drupal uses the IETF BCP 47 standard which uses more
+        // characters. For example this will convert 'pt-pt' to 'pt'.
+        'exclude' => [substr($this->languageManager->getCurrentLanguage()->getId(), 0, 2)],
       ],
       'renderAs' => $render_as_options,
       'domain' => $this->configuration['domain'],
--- a/modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
+++ b/modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
@@ -5,6 +5,7 @@ declare(strict_types = 1);
 namespace Drupal\oe_webtools_etrans\Plugin\Block;
 
 use Drupal\Component\Serialization\Json;
+use Drupal\Component\Utility\Html;
 use Drupal\Core\Block\BlockBase;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Language\LanguageInterface;
@@ -27,6 +28,14 @@ class ETransBlock extends BlockBase implements ContainerFactoryPluginInterface {
    * The various ways the eTrans link can be rendered.
    */
   public const RENDER_OPTIONS = ['button', 'icon', 'link'];
+
+  /**
+   * Default values for the configuration options of this block.
+   */
+  protected const DEFAULT_CONFIGURATION = [
+    'render_as' => 'button',
+    'render_to' => '',
+  ];
 
   /**
    * The language manager.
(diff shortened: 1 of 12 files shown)


modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
```diff
@@ -0,0 +1,132 @@
...
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, LanguageManagerInterface $language_manager) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->languageManager = $language_manager;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
```
claudiu-cristea, line 58: Let's add return strict typing.

modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
```diff
@@ -0,0 +1,132 @@
...
+      $plugin_id,
+      $plugin_definition,
+      $container->get('language_manager')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
```
claudiu-cristea, line 70: Return strict typing?

modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
```diff
@@ -0,0 +1,132 @@
...
+    ];
+
+    return $build;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function defaultConfiguration() {
+    return parent::defaultConfiguration() + [
```
claudiu-cristea, line 100: This is only theoretical but we normally override in implementation the parent's configuration.  In this way an unlikely parent `render_as` can be overridden:

```php
return [
  'render_as' => 'button',
] + parent::defaultConfiguration();
```

modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
```diff
@@ -0,0 +1,132 @@
...
+      '#attributes' => ['type' => 'application/json'],
+    ];
+
+    return $build;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function defaultConfiguration() {
```
claudiu-cristea, line 99: Missing return strict typing.

modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
```diff
@@ -0,0 +1,132 @@
...
+    return parent::defaultConfiguration() + [
+      'render_as' => 'button',
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
```
claudiu-cristea, line 108: Same,

modules/oe_webtools_etrans/tests/src/Kernel/ETransBlockTest.php
```diff
@@ -0,0 +1,113 @@
...
+namespace Drupal\Tests\oe_webtools_etrans\Kernel;
+
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\language\Entity\ConfigurableLanguage;
+use Drupal\oe_webtools_etrans\Plugin\Block\ETransBlock;
+
+/**
+ * Tests the block that displays the Webtools eTrans link.
+ */
+class ETransBlockTest extends KernelTestBase {
```
claudiu-cristea, line 14: I think checking the rendered script tag, as is done here, is enough. Webtools should provide the functional testing.

modules/oe_webtools_etrans/tests/src/Kernel/ETransBlockTest.php
```diff
@@ -0,0 +1,113 @@
...
+   * The service to set the default language.
+   *
+   * @var \Drupal\Core\Language\LanguageDefault
+   */
+  protected $languageDefault;
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp() {
```
claudiu-cristea, line 44: Deserves an return type.

claudiu-cristea (changes requested):
The PR looks good and I've already had the chance to test it on Joinup. The implementation of webtools widget is straight and follows the other `oe_webtools_*` modules patterns. I only have some nits.

modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
pfrenssen: Oh wow yeah this is a bug, thanks!

modules/oe_webtools_etrans/tests/src/Kernel/ETransBlockTest.php
pfrenssen: I agree, this is the limit of our responsibility. This is also consistent with the other modules in `oe_webtools`.

pfrenssen:
Thanks for the review! I agree with all the points, all have been addressed.

pfrenssen:
We added a new test in Joinup yesterday that validates the schema of all exported config. It uncovered that the config schema was missing for the block that is added in this PR. I rectified this.

claudiu-cristea (dismissed):

pfrenssen:
We are now planning to add some additional options to the block, putting this back in draft.

modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
```diff
@@ -0,0 +1,226 @@
...
+
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function blockValidate($form, FormStateInterface $form_state) {
+    $render_to_value = $form_state->getValue('render_to');
+    if (!empty($render_to_value) && $render_to_value !== Html::cleanCssIdentifier($render_to_value)) {
```
claudiu-cristea, line 208: Nice 👍

claudiu-cristea (approved):
Now it has all options supported by eTrans. Thank you!

22Alexandra (dismissed):

modules/oe_webtools_etrans/tests/src/Kernel/ETransBlockTest.php
```diff
@@ -0,0 +1,170 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+    $this->installEntitySchema('user');
+    $this->installSchema('system', 'sequences');
+    $this->installConfig(['system']);
+    $this->languageManager = \Drupal::service('language_manager');
```
yenyasinn, line 48:
```suggestion
    $this->languageManager =$this->container->get('language_manager');
    $this->languageDefault = $this->container->get('language.default');
```
There is big discussion about how services have to be invoked in tests https://www.drupal.org/project/drupal/issues/2066993. For Kernel test usage of $this->container is safe and this approach is used in the test below. So let's do it consistent in whole test, if you don't have objections.

pfrenssen: Please read again carefully the issue you linked. This is exactly why I am using `\Drupal::service()` instead of `$this->container()`.

I see now that the issue summary is not correct but this is dating from 8 years. The discussion in this issue has done a complete 180 degree flip against the original proposal, and is now advocating for the opposite. The relevant discussion starts at comment 35 (written by Berdir). This and subsequent discussion is advocating _against_ the use of `$this->container`, giving a lot of evidence and examples of problems arising from it.

So if possible I would prefer to keep it, it is the current best practice AFAIK.

yenyasinn: Yes, sure it can be kept as is.
 I've read whole thread. It was mentioned that $this->container shouldn't be used in functional test but can be used in kernel test. My only concern is that \Drupal::service() and $this->container() are used both in this test.

pfrenssen: You're right, I saw too that most comments are actually for `BrowserTestBase`, Kernel tests are safe, just as you said.

I will approve this suggestion. Thanks for catching it!

modules/oe_webtools_etrans/tests/src/Kernel/ETransBlockTest.php
```diff
@@ -0,0 +1,171 @@
...
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+    $this->installEntitySchema('user');
+    $this->installSchema('system', 'sequences');
+    $this->installConfig(['system']);
+    $this->languageManager = $this->container->get('language_manager');
+    $this->languageDefault = $this->container->get('language.default');
+    $this->languageDefault = \Drupal::service('language.default');
```
yenyasinn, line 50:
```suggestion
```
We don't need this line anymore.

yenyasinn (dismissed):

pfrenssen:
@claudiu-cristea there were a few changes after your last review, if you have some time could you check these and re-approve? Thanks!

claudiu-cristea (dismissed):
Approving the last changes.

idimopoulos (dismissed):

Joaoeduardoramos:
We are testing this PR at efsa too and it looks good, works as expected and is very useful. So a big +1 for this. Thank you !

nagyad (dismissed):

yenyasinn (dismissed):

modules/oe_webtools_etrans/src/Plugin/Block/ETransBlock.php
```diff
@@ -0,0 +1,226 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function build(): array {
+    $render_as_options = array_fill_keys(self::RENDER_OPTIONS, FALSE);
+    $render_as_options[$this->configuration['render_as']] = TRUE;
+    $json = [
+      'service' => 'etrans',
+      'languages' => [
+        'exclude' => [$this->languageManager->getCurrentLanguage()->getId()],
```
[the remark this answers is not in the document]
upchuk: Yes, but any site that uses this and will use the PT language will have a problem with this.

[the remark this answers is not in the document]
claudiu-cristea: @sergepavle What exactly is the _pt-pt issue_?

sergepavle (approved):

PR #149, opened 2021-01-22
ISAICP-6273: Avoid CORS issues by requesting the Webtools Smartloader script over HTTPS

### LINKED ISSUE #148: Load the webtools script over HTTPS
When working on a local development environment the various Webtools JS scripts are rejected because they are cross-origin requests.

According to the Webtools [Guidelines for developers](https://webgate.ec.europa.eu/fpfis/wikis/display/webtools/Important+guidelines+for+developers+and+implementers) this can be solved by requesting the main JS loader using HTTPS, and setting the local hostname correctly.

> **Always use HTTPS**
> Even if the Webtools Smart Loader is protocol agnostic, in order to prevent any CORS issues and mismatch of protocols, reference the Webtools Smart Loader using HTTPS.

I can confirm that my CORS issues magically disappeared when I request `load.js` over HTTPS rather than over HTTP.

pfrenssen:
Fixes #148

### Description

According to the Webtools documentation this resource should be accessed over HTTPS.

Ref. https://webgate.ec.europa.eu/fpfis/wikis/display/webtools/Important+guidelines+for+developers+and+implementers

### Change log

- Changed: The Webtools Smartloader script is requested over HTTPS.

### DIFF
--- a/oe_webtools.libraries.yml
+++ b/oe_webtools.libraries.yml
@@ -2,4 +2,4 @@
 drupal.webtools-smartloader:
   version: 1.x
   js:
-    "https://europa.eu/webtools/load.js": { external: true, attributes: { defer: true } }
+    https://europa.eu/webtools/load.js: { type: external, attributes: { defer: true } }
--- a/oe_webtools.libraries.yml
+++ b/oe_webtools.libraries.yml
@@ -2,4 +2,4 @@
 drupal.webtools-smartloader:
   version: 1.x
   js:
-   //europa.eu/webtools/load.js: { external: true, attributes: { defer: true } }
+    "https://europa.eu/webtools/load.js": { external: true, attributes: { defer: true } }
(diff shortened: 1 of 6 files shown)


oe_webtools.libraries.yml
```diff
@@ -2,4 +2,4 @@
 drupal.webtools-smartloader:
   version: 1.x
   js:
-   //europa.eu/webtools/load.js: { external: true, attributes: { defer: true } }
+    "https://europa.eu/webtools/load.js": { external: true, attributes: { defer: true } }
```
brummbar, line 5: We noticed that `external: true` is wrong, it's actually `type: external` 😅  
Also very nit but the quotes can be avoided in this case.

brummbar (changes requested):

brummbar (approved):

sergepavle (commented):
Everything looks good. It would be nice to update also:
tests/fixtures/static_html/one_map.html
tests/fixtures/static_html/two_map.html
It doesn't have an impact on anything but it would be nice to update.

brummbar (approved):

sergepavle (approved):

PR #152, opened 2021-03-17
EWPP-842: Allow chart and racing services for webtols chart widget.

22Alexandra:


sergepavle (dismissed):

modules/oe_webtools_media/src/Plugin/media/Source/Webtools.php
```diff
@@ -124,22 +124,22 @@ public function getWidgetTypes() {
     return [
       'chart' => [
         'name' => $this->t('Chart'),
-        'service' => 'charts',
+        'services' => ['charts', 'chart', 'racing'],
```
upchuk, line 127: This method is a public API, and hence, breaks BC. So we need to leave `service` in place with `charts`. If people use this and rely on `service`, we should not break that. And if they want both, they can update their code.

modules/oe_webtools_media/src/Plugin/Validation/Constraint/WebtoolsMediaConstraintValidator.php
```diff
@@ -46,7 +46,7 @@ public function validate($value, Constraint $constraint) {
     }
 
     // Add violation in case incorrect services.
-    if (!isset($snippet['service']) || $snippet['service'] !== $widget_types[$constraint->widgetType]['service']) {
+    if (!isset($snippet['service']) || !in_array($snippet['service'], $widget_types[$constraint->widgetType]['services'])) {
```
upchuk, line 49: Also BC issue here. Check for the presence of `services` and if not, case `service` to an array and use that. This is in case someone extended the source class.

upchuk (dismissed):

modules/oe_webtools_media/src/Plugin/media/Source/Webtools.php
yenyasinn: @upchuk, should we call `service` parameter deprecated to enforce update it by 3rd parties and don't support the same parameters?

upchuk: Ah yes yes, i wanted to say that as well but forgot.

modules/oe_webtools_media/src/Plugin/media/Source/Webtools.php
```diff
@@ -125,6 +125,7 @@ public function getWidgetTypes() {
       'chart' => [
         'name' => $this->t('Chart'),
         'service' => 'charts',
+        'services' => ['chart', 'racing'],
```
yenyasinn, line 128: If `service` is deprecated then need to provide `services` parameter for each widget type as was done in the first commits.

upchuk: yes, all of them need to get `services`

sergepavle (dismissed):

yenyasinn (dismissed):

modules/oe_webtools_media/src/Plugin/media/Source/Webtools.php
```diff
@@ -124,22 +124,30 @@ public function getWidgetTypes() {
     return [
       'chart' => [
         'name' => $this->t('Chart'),
-        'service' => 'charts',
+        // @deprecated Use services key instead.
```
upchuk, line 127: This is not good. My comment said: ` So we need to leave service in place with charts`. You removed `charts` hence broke BC...

upchuk (changes requested):

upchuk (approved):

sergepavle (approved):

PR #164, opened 2021-06-15
EWPP-1178: Update globan to use the new json syntax.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


22Alexandra (approved):

tests/Behat/WebtoolsGlobanContext.php
```diff
@@ -46,19 +47,19 @@ public function backupGlobanConfigs(): void {
...
-   * @param string|null $globan_lang
-   *   The globan language option value.
+   * @param string $value
+   *   The expected JSON string.
    *
-   * @Then the Webtools javascript is loaded with the globan options :globan_option
-   * @Then the Webtools javascript is loaded with the globan options :globan_option and language :globan_lang
+   * @throws \Exception
+   *
+   * @Then the page should have globan json snippet :value
```
brummbar, line 57: I think we can finally improve this, fetch the script that contains `"utility":"globan"`, read its values and compare them with the ones provided in a table:
```
Then the Webtools globan should be loaded with the values: 
| name | value |
| theme | dark |
| logo | false |
|...|
```

What do you think?

brummbar (changes requested):

tests/Behat/WebtoolsGlobanContext.php
sergepavle: I had this idea already but currently, I already noticed that behat always pass from behat step data as string including from table. JSON and parsed data more strict according to data type. For sure we can massage data from behat or JSON to some unified data type (convert all values to string (if JSON data) or cast values from behat Table.
But it looks a bit more tricky and anyway more accurate to compare full json string.

brummbar: True, it would be more hacky. Let's keep it this way.

modules/oe_webtools_globan/src/Form/WebtoolsGlobanSettingsForm.php
```diff
@@ -117,6 +117,13 @@ public function buildForm(array $form, FormStateInterface $form_state): array {
       '#default_value' => $config->get('sticky') ? 1 : 0,
     ];
 
+    $form['globan_settings']['zindex'] = [
+      '#type' => 'number',
+      '#title' => $this->t('Z-index'),
+      '#description' => $this->t('Makes the banner z-index value depending on your layout design. Default is 40.'),
```
brummbar, line 123:
```suggestion
      '#description' => $this->t('Adapt the banner z-index value depending on your layout design. Default is 40.'),
```

modules/oe_webtools_globan/oe_webtools_globan.module
```diff
@@ -7,68 +7,47 @@
...
+use Drupal\Component\Serialization\Json;
 use Drupal\Core\Cache\CacheableMetadata;
-use Drupal\Core\Url;
 
 /**
- * Implements hook_page_attachments().
+ * Implements hook_page_top().
  *
- * Attach smart loader.
+ * Attach smart loader with appropriate JSON snippet for the global banner.
```
brummbar, line 16: Let's add a comment here: We inject the globan JSON snippet in the page top area because the banner markup is injected in the same HTML element that contains the JSON. Because of this, we cannot place the JSON in the page head element.

brummbar (changes requested):
All the rest looks good.

brummbar (approved):

nagyad (approved):

brummbar (approved):

nagyad (approved):

PR #168, opened 2021-08-18
[EWPP-1424] EWPP-1425: Adds configuration form for the Webtools Page Feedback module.

22Alexandra:


modules/oe_webtools_page_feedback/README.md
```diff
@@ -0,0 +1,29 @@
...
+* Enabled: Whether the Page Feedback Form is enabled or not.
+* Form ID: The Webtools ID of the corresponding Feedback Form where submissions will be recorded.
+
+This configuration can be provided using the configuration form (`/admin/config/system/oe_webtools_page_feedback`) by
+users that were granted with the permission `administer webtools page feedback form` or by providing details in your
+site's settings.php file:
+
+```
+$config['oe_webtools_page_feedback.settings']['enabled'] = TRUE;
+$config['oe_webtools_page_feedback.settings']['formID'] = '1234';
```
sergepavle, line 28: Usually, we don't use camel cases in elements of the configuration.
Using `form_id` probably would be more consistent.

sergepavle (approved):

modules/oe_webtools_page_feedback/README.md
sergepavle: But in general, it is fine for me to have this variant as well.

modules/oe_webtools_page_feedback/src/Form/WebtoolsPageFeedbackSettingsForm.php
```diff
@@ -0,0 +1,63 @@
...
+      '#description' => $this->t('Check this box if you would like to enable the Page feedback form on all the content pages.'),
+      '#default_value' => $config->get('enabled'),
+    ];
+    $form['formID'] = [
+      '#type' => 'number',
+      '#title' => $this->t('Form ID'),
+      '#description' => $this->t('Provide your webtools form ID.'),
+      '#default_value' => $config->get('formID'),
+      '#required' => TRUE,
+    ];
```
sergepavle, line 40: To have this element wrapped in fieldset would be a bit more `nicer`

modules/oe_webtools_page_feedback/README.md
imanoleguskiza: Lets make this change, I agree with Sergei in this case.

modules/oe_webtools_page_feedback/tests/src/Functional/PageFeedbackFormTest.php
```diff
@@ -0,0 +1,51 @@
...
+    // Assert default values.
+    $this->assertSession()->pageTextContains('Webtools Page Feedback Form settings');
+    $this->assertSession()->checkboxNotChecked('Enabled');
+    $this->assertSession()->fieldValueEquals('Form ID', '');
+    // Configure the form.
+    $page = $this->getSession()->getPage();
+    $page->checkField('Enabled');
+    $page->fillField('Form ID', '1234');
+    $page->pressButton('Save configuration');
+    // Assert values are correctly saved.
```
imanoleguskiza, line 44: Lets just add a quick assertion to check that the form shows the stored values as well.

modules/oe_webtools_page_feedback/src/Form/WebtoolsPageFeedbackSettingsForm.php
```diff
@@ -0,0 +1,63 @@
...
+    $form['enabled'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Enabled'),
+      '#description' => $this->t('Check this box if you would like to enable the Page feedback form on all the content pages.'),
+      '#default_value' => $config->get('enabled'),
+    ];
+    $form['formID'] = [
+      '#type' => 'number',
+      '#title' => $this->t('Form ID'),
+      '#description' => $this->t('Provide your webtools form ID.'),
```
imanoleguskiza, line 37: I'm wondering if we shouldn't provide more info here on how to request the ID... Let's leave it like this and we can fix it if we need later in the epic.

imanoleguskiza (changes requested):

modules/oe_webtools_page_feedback/src/Form/WebtoolsPageFeedbackSettingsForm.php
22Alexandra: This is how you request a service https://webgate.ec.europa.eu/fpfis/wikis/display/webtools/Default+Feedback+Form#DefaultFeedbackForm-Howtorequestthisservice and there is a link in the [README](https://github.com/openeuropa/oe_webtools/pull/168/files#diff-cfffa78681c6bd04aee6f8351dd5c3852589cdfb8ab5b1c791768fe0524ce886R9) pointing to the webtools widget page so developers can find the information. For ewcms only CEM should have access (they would know already the form id).

Do you think it's still necessary?

imanoleguskiza: Good point, CEM should already be aware of this so lets leave it like this.

imanoleguskiza (approved):

modules/oe_webtools_page_feedback/README.md
```diff
@@ -0,0 +1,29 @@
+# OpenEuropa Webtools Page Feedback
+
+Provides a widget that integrates with the Webtools Default Feedback Form allowing the users to report on the page
+usefulness and potentially provide suggestions to improve the content.
+
+### How it works
+
+This module outputs the Webtools Default Feedback Form on all the content types when enabled.
```
upchuk, line 8: No it doesnt. This module simply provides a block that does that no?

modules/oe_webtools_page_feedback/config/schema/oe_webtools_page_feedback.schema.yml
```diff
@@ -0,0 +1,12 @@
+oe_webtools_page_feedback.settings:
+  type: config_object
+  label: 'Webtools Page Feedback Form settings'
+  mapping:
+    enabled:
+      type: boolean
+      label: 'Enabled'
+      description: 'Whether the Page Feedback Form is enabled or not.'
+    feedback_form_id:
+      type: string
```
upchuk, line 10: The reqs say integer no?

modules/oe_webtools_page_feedback/src/Form/WebtoolsPageFeedbackSettingsForm.php
```diff
@@ -0,0 +1,63 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state): array {
+    $config = $this->config('oe_webtools_page_feedback.settings');
+
+    $form['enabled'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Enabled'),
+      '#description' => $this->t('Check this box if you would like to enable the Page feedback form on all the content pages.'),
```
upchuk, line 31: I was wrong here, let's say only: "Check this box if you would like to enable the Page feedback form on this site."

modules/oe_webtools_page_feedback/src/Form/WebtoolsPageFeedbackSettingsForm.php
upchuk: We are in a component here, so CEM is besides the point. But yea, we link in the readme to webtools so that should be fine.

modules/oe_webtools_page_feedback/src/Form/WebtoolsPageFeedbackSettingsForm.php
```diff
@@ -0,0 +1,63 @@
...
+      '#default_value' => $config->get('enabled'),
+    ];
+    $form['feedback_form_id'] = [
+      '#type' => 'number',
+      '#title' => $this->t('Form ID'),
+      '#description' => $this->t('Provide your webtools form ID.'),
+      '#default_value' => $config->get('feedback_form_id'),
+      '#required' => TRUE,
```
upchuk, line 39: Hmm, this should only be required if the thing is enabled right? So if they first enable and set a form ID, then they need to leave the form ID there if they wanna disable. So let's add a #state here for that.

upchuk (changes requested):

sergepavle (approved):

PR #169, opened 2021-08-19
[EWPP-1424] EWPP-1426: Provide a block that renders the DFF webtools widget.

22Alexandra:


modules/oe_webtools_page_feedback/oe_webtools_page_feedback.module
```diff
@@ -0,0 +1,32 @@
...
+declare(strict_types = 1);
+
+use Drupal\Core\Cache\CacheableMetadata;
+
+/**
+ * Implements hook_page_attachments().
+ *
+ * Attach smart loader.
+ */
+function oe_webtools_page_feedback_page_attachments(array &$attachments): void {
```
upchuk, line 17: Why you attaching this here? Just make add it to the the render array in the block where we render the actual thing...like SocialShareBlock. Which in fact ..you do!

modules/oe_webtools_page_feedback/src/EventSubscriber/PageFeedbackConfigCacheInvalidator.php
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Invalidate cache tags when page feedback form config object changes.
+   *
+   * @param \Drupal\Core\Config\ConfigCrudEvent $event
+   *   The event to process.
+   */
+  public function onChange(ConfigCrudEvent $event) {
+    if ($event->getConfig()->getName() === 'oe_webtools_page_feedback.settings') {
+      $this->cacheTagsInvalidator->invalidateTags(['library_info', 'rendered']);
```
upchuk, line 42: Why is this needed? We should just add the cache tags on the block render array. Saving the config should not add/remove a library from the registry so I fail to see why this is needed.

modules/oe_webtools_page_feedback/src/Plugin/Block/PageFeedbackFormBlock.php
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build(): array {
+    $feedback_form_json = [
+      'service' => 'dff',
+      'id' => $this->configFactory->get('oe_webtools_page_feedback.settings')->get('feedback_form_id'),
+      'lang' => $this->languageManager->getCurrentLanguage()->getId(),
```
upchuk, line 94: Does this work with PT? As far as I know, they expect PT, not pt-pt.

modules/oe_webtools_page_feedback/tests/src/FunctionalJavascript/PageFeedbackFormTest.php
```diff
@@ -24,9 +28,23 @@ class PageFeedbackFormTest extends WebDriverTestBase {
...
+    parent::setUp();
+
+    $this->drupalPlaceBlock('oe_webtools_page_feedback_form');
+    $this->drupalPlaceBlock('page_title_block');
+    $this->drupalCreateContentType(['type' => 'page', 'name' => 'Basic page']);
+    ConfigurableLanguage::createFromLangcode('pt')->save();
+  }
+
+  /**
+   * Tests Page Feedback Form configuration and block rendering.
```
upchuk, line 43: Then the class should also be renamed no?

modules/oe_webtools_page_feedback/tests/src/FunctionalJavascript/PageFeedbackFormTest.php
```diff
@@ -24,9 +28,23 @@ class PageFeedbackFormTest extends WebDriverTestBase {
...
-   * Tests Page Feedback Form configuration.
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+
+    $this->drupalPlaceBlock('oe_webtools_page_feedback_form');
+    $this->drupalPlaceBlock('page_title_block');
+    $this->drupalCreateContentType(['type' => 'page', 'name' => 'Basic page']);
+    ConfigurableLanguage::createFromLangcode('pt')->save();
```
upchuk, line 39: What language is `pt`? Because the Drupal language code for Portuguese is `pt-pt`.

upchuk (changes requested):

modules/oe_webtools_page_feedback/oe_webtools_page_feedback.module
sergepavle: Fixed.

modules/oe_webtools_page_feedback/src/EventSubscriber/PageFeedbackConfigCacheInvalidator.php
sergepavle: Removed.

modules/oe_webtools_page_feedback/src/Plugin/Block/PageFeedbackFormBlock.php
sergepavle: Webtools don't support pt-pt lang code. Seems like the issue is related not only to this widget but also to other submodules. Should we do a follow-up regarding these issues?

modules/oe_webtools_page_feedback/tests/src/FunctionalJavascript/PageFeedbackFormTest.php
sergepavle: done

modules/oe_webtools_page_feedback/tests/src/FunctionalJavascript/PageFeedbackFormTest.php
sergepavle: fixed

upchuk (changes requested):

PR #177, opened 2022-02-01
EWPP-1910: Update to PHP7.4/8.0

hernani:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,9 +8,14 @@
     "require": {
         "php": ">=7.4",
         "ext-json": "*",
-        "drupal/core": "^9.2"
+        "drupal/core": "^9.2",
+        "drupal/ctools": "^3.7",
+        "drupal/file_link": "2.0.4",
+        "drupal/pathauto": "^1.8",
+        "drupal/token": "^1.10"
     },
     "require-dev": {
+        "behat/behat": "^3.10",
         "composer/installers": "~1.5",
         "drupal/core-composer-scaffold": "^9.2",
         "drupal/core-dev": "^9.2",
@@ -20,6 +25,7 @@
         "drupal/geofield": "^1.0",
         "drupal/json_field": "~1.0-rc4",
         "drush/drush": "~10.3",
+        "egulias/email-validator": "^2.1.22 || ^3.0",
         "nikic/php-parser": "^4.12.0",
         "drupaltest/behat-traits": "dev-GH-10",
         "openeuropa/behat-transformation-context": "~0.1",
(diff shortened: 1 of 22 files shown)


composer.json
```diff
@@ -6,30 +6,37 @@
...
-        "drupaltest/behat-traits": "dev-GH-6",
+        "egulias/email-validator": "^2.1.22 || ^3.0",
+        "nikic/php-parser": "^4.12.0",
+        "drupaltest/behat-traits": "dev-GH-10",
         "openeuropa/behat-transformation-context": "~0.1",
-        "openeuropa/code-review": "~1.6",
-        "openeuropa/drupal-core-require-dev": "^8.9 || ^9",
+        "openeuropa/code-review": "^2.0",
         "openeuropa/oe_media": "^1.14",
         "openeuropa/oe_multilingual": "^1.8",
```
sergepavle, line 34: We have to use ^1.10 which is compatible with PHP8.

composer.json
```diff
@@ -6,30 +6,37 @@
...
     "require": {
-        "php": ">=7.3",
+        "php": ">=7.4",
         "ext-json": "*",
-        "drupal/core": "^8.9 || ^9"
+        "drupal/core": "^9.2",
+        "drupal/ctools": "^3.7",
+        "drupal/file_link": "2.0.4",
+        "drupal/pathauto": "^1.8",
+        "drupal/token": "^1.10"
```
sergepavle, line 15: I've found that all these dependencies coming from oe_multilingual which is from `require-dev`. Why we cannot put them within require-dev as well?

composer.json
```diff
@@ -6,30 +6,37 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "require": {
-        "php": ">=7.3",
+        "php": ">=7.4",
         "ext-json": "*",
-        "drupal/core": "^8.9 || ^9"
+        "drupal/core": "^9.2",
+        "drupal/ctools": "^3.7",
+        "drupal/file_link": "2.0.4",
```
sergepavle, line 13: Why do we need to lock the file_link version? I've tried locally and with `^2.0.4` it works fine as well.

composer.json
```diff
@@ -6,30 +6,37 @@
...
+        "drupal/pathauto": "^1.8",
```
sergepavle, line 14: After setting oe_multilingual minimal version to 1.10 didn't see the need to set pathauto minimum version.

composer.json
```diff
@@ -6,30 +6,37 @@
...
         "drupal/config_devel": "~1.2",
         "drupal/drupal-extension": "^4.0",
         "drupal/geocoder": "^3.0",
         "drupal/geofield": "^1.0",
         "drupal/json_field": "~1.0-rc4",
         "drush/drush": "~10.3",
-        "nikic/php-parser": "^3.1.5",
-        "drupaltest/behat-traits": "dev-GH-6",
+        "egulias/email-validator": "^2.1.22 || ^3.0",
+        "nikic/php-parser": "^4.12.0",
```
sergepavle, line 29: Do we still need this dependency?

sergepavle (changes requested):

composer.json
22Alexandra: We do need as long as we support PHP7.4.

sergepavle (dismissed):

nagyad (dismissed):

22Alexandra (approved):

sergepavle (approved):

PR #182, opened 2022-02-19
OEL-1996: Fix PHP 8.1 deprecation but keep PHP 7.4 compatibility

claudiu-cristea:
### Description

With PHP 8.1, we get this message:

> Return type of Drupal\oe_webtools_analytics\Event\AnalyticsEvent::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice

### Change log

- Added: `#[\ReturnTypeWillChange]` to `AnalyticsEvent::jsonSerialize()`. Note, that we cannot change the return typing as we still support PHP 7.4 but the mixed pseudo type was added only on PHP 8.0.
- Changed: None.
- Deprecated: None.
- Removed: None.
- Fixed: None.
- Security: None.

### Commands

n/a

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,6 @@
     "require-dev": {
         "behat/behat": "^3.10",
         "composer/installers": "~1.5",
-        "drupal/coder": "^8.3.15",
         "drupal/config_devel": "~1.2",
         "drupal/core-composer-scaffold": "^9.3",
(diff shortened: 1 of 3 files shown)


claudiu-cristea:
This was fixed already in `master`. Closing.

PR #184, opened 2022-03-21
EWPP-2085: Allow alpha numeric site id for Webtools analytics.

22Alexandra:


nagyad (dismissed):

sergepavle (approved):

modules/oe_webtools_analytics/tests/src/Functional/ConfigurationTest.php
```diff
@@ -32,13 +31,16 @@ class ConfigurationTest extends BrowserTestBase {
...
-      ->set('instance', 'testing');
-    $config->save();
+    $user = $this->createUser(['administer webtools analytics']);
+    $this->drupalLogin($user);
+    $this->drupalGet('admin/config/system/oe_webtools_analytics');
+    $page = $this->getSession()->getPage();
+    $page->fillField('Site ID', '123');
+    $page->fillField('Site path', 'ec.europa.eu');
+    $page->fillField('Instance', 'testing');
+    $page->pressButton('Save configuration');
```
sergepavle, line 41: Why are we changing from direct modifying config to update of config through UI?
I thought it was fine as before. Anyway, we are changing it through UI later in this test.

22Alexandra: Just making sure the field can be filled in both with digits only and alphanumeric values.

nagyad (approved):

Issue #203, opened 2022-11-22
Webtools Sliding Captcha integration


cor-emble:
Hi all,

I've created a small custom module for a EC project we maintain which implements the Webtools Sliding captcha on Drupal forms. 

It is build on the Captcha API module https://www.drupal.org/project/captcha so you could add the captcha on all forms supported by the Captcha module like Webforms. 

Would love to share it but should this be part of the webtools module or should we create a new repository?

brummbar:
Hello @cor-emble , that sounds incredibly nice! Can you open a pull request with the code? If it's well isolated it can be a submodule.

PR #218, opened 2023-07-07
EWPP-3386: Implement the webtools sliding captcha.

nagyad:
Will resolve https://github.com/openeuropa/oe_webtools/issues/203.


modules/oe_webtools_captcha/config/install/oe_webtools_captcha.settings.yml
```diff
@@ -0,0 +1 @@
+validationEndpoint: 'https://europa.eu/webtools/rest/captcha/verify'
```
sergepavle, line 1: As I understand from the initial plans in the ticket it should be configured through settings, not config?

modules/oe_webtools_captcha/oe_webtools_captcha.module
```diff
@@ -0,0 +1,94 @@
...
+  }
+  return $captcha;
+}
+
+/**
+ * Validate the captcha answer against the validation API.
+ */
+function oe_webtools_captcha_captcha_validation($solution, $response, $element, $form_state): bool {
+  $validation_endpoint = \Drupal::config('oe_webtools_captcha.settings')->get('validationEndpoint');
+  $url = Url::fromUri($validation_endpoint);
```
sergepavle, line 53: Better probably to wrap with `try` to avoid fatal errors with incorrect URL.

modules/oe_webtools_captcha/tests/modules/oe_webtools_captcha_mock/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+```bash
+./vendor/bin/drush en oe_webtools_captcha_mock
+```
+Please note that it will set the mock route as the validation endpoint during
+installation!
+
+### Usage
+
+Drupal state variable `captcha_mock_response` is used to determine the
```
sergepavle, line 15:
```suggestion
Drupal state variable `captcha_mock_response` is used to determine
```

sergepavle (changes requested):

modules/oe_webtools_captcha/config/install/oe_webtools_captcha.settings.yml
nagyad: I discussed it with Hernani, it's be fine.

modules/oe_webtools_captcha/oe_webtools_captcha.module
```diff
@@ -0,0 +1,94 @@
...
+ */
+function oe_webtools_captcha_captcha($op, $captcha_type = '') {
+  $captcha = [];
+  switch ($op) {
+    case 'list':
+      return ['Webtools captcha'];
+
+    case 'generate':
+      if ($captcha_type === 'Webtools captcha') {
+        $captcha['solution'] = TRUE;
```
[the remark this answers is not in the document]
nagyad: `Note that if this TRUE, the 'captcha_validate' key must be set to a callback that ignores the solution.` 
I'm not sure this means exactly what you mean.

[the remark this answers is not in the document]
sergepavle: yes. you're right. 'solution' value shouldn't be empty.

sergepavle (approved):

modules/oe_webtools_captcha/tests/modules/oe_webtools_captcha_mock/src/Controller/MockController.php
```diff
@@ -0,0 +1,66 @@
...
+    );
+  }
+
+  /**
+   * Mock validation.
+   *
+   * @return \Symfony\Component\HttpFoundation\JsonResponse
+   *   A JSON response.
+   */
+  public function validate(): JsonResponse {
```
22Alexandra, line 49: If we are mocking the validation we don't test the Webtools response but a hardcoded one from this controller.
Checking the [Webtools captcha wiki](https://webgate.ec.europa.eu/fpfis/wikis/pages/viewpage.action?spaceKey=webtools&title=Sliding+Captcha) there are 2 parameters `wt_captcha_answer` and `wt_captcha_sid` that should be used in order to validate the response. Can we do something similar to `MockWcloudMiddleware` and assert all the cases instead of hardcoded responses?

22Alexandra (changes requested):

modules/oe_webtools_captcha/tests/modules/oe_webtools_captcha_mock/src/Controller/MockController.php
22Alexandra: Checked with @hernani, testing the response is handled should be enough. No need for testing the actual Webtools response.

hernani (dismissed):

hernani (approved):

sergepavle (approved):

22Alexandra (approved):

PR #221, opened 2023-09-26
EWPP-3623: Add Webtools Wtag service.

22Alexandra:


oe_webtools.libraries.yml
```diff
@@ -3,3 +3,8 @@ drupal.webtools-smartloader:
   version: 1.x
   js:
     https://europa.eu/webtools/load.js: { type: external, attributes: { defer: true } }
+
+oe_webtools.component-collect:
```
imanoleguskiza, line 7: We don't seem to be using this library?

modules/oe_webtools_wtag/src/Plugin/Field/FieldWidget/WtagWidget.php
```diff
@@ -0,0 +1,125 @@
...
+   */
+  public static function isApplicable(FieldDefinitionInterface $field_definition) {
+    // This widget is only applicable to unlimited cardinality SKOS Concept
+    // reference fields that reference the DET vocabulary.
+    if ($field_definition->getFieldStorageDefinition()->getCardinality() !== FieldStorageDefinitionInterface::CARDINALITY_UNLIMITED) {
+      return FALSE;
+    }
+
+    $concept_schemes = $field_definition->getSetting('handler_settings')['concept_schemes'];
+    if (!in_array('http://data.europa.eu/uxp/det', $concept_schemes)) {
```
imanoleguskiza, line 81: Should we allow to use this widget for a field that is configured to use the DET vocabulary AND some other vocabulary? It seems weird as it essentially prevents the user from choosing anything but the det vocabulary

modules/oe_webtools_wtag/src/Element/Wtag.php
```diff
@@ -0,0 +1,135 @@
...
+      'service' => 'wtag',
+      'target' => '#' . $element['#id'],
+      'title' => $element['#modal_title'],
+      'description' => $element['#modal_description'],
+    ];
+
+    $element['#wtag'] = Markup::create(Json::encode($wtag_json));
+    $element['#attached']['library'][] = 'oe_webtools/drupal.webtools-smartloader';
+    unset($element['#attributes']['data-autocomplete-path']);
+    $element['#attributes']['maxlength'] = '100000000';
```
imanoleguskiza, line 66: Are both maxlengths needed?

imanoleguskiza (changes requested):

modules/oe_webtools_wtag/src/Plugin/Field/FieldWidget/WtagWidget.php
22Alexandra: The Webtools Wtag widget only allows DET, so I don't think the user would be able to choose terms from other vocabularies.

oe_webtools.libraries.yml
upchuk: @imanoleguskiza we will in EWCMS, and it's generic. Used in cases when we need to add aa webtools widget as part of an ajax request.

modules/oe_webtools_wtag/src/Plugin/Field/FieldWidget/WtagWidget.php
upchuk: @imanoleguskiza They only built support for DET, quite hardcoded on their end as well..

modules/oe_webtools_wtag/src/Element/Wtag.php
upchuk: Well, I think so, don't remember 100%. One is for the HTML element itself, the other is checked by a validator somewhere..

modules/oe_webtools_wtag/src/Plugin/Field/FieldWidget/WtagWidget.php
upchuk: @imanoleguskiza Yes, true.
@22Alexandra could you please update this to ensure that there is a single concept scheme selected and that it is DET? Thanks

imanoleguskiza (approved):

imanoleguskiza (approved):

upchuk (approved):

PR #245, opened 2024-07-26
OEL-3156: Placed widgets inside the body tag

drishu:
### Description

All widgets ("application/json") should be placed inside the body tag as described in webtools documentation: https://webgate.ec.europa.eu/fpfis/wikis/display/webtools/Widgets

### Change log

- Changed: From hook_page_attachments to hook_page_bottom

### DIFF
--- a/modules/oe_webtools_analytics/tests/src/Functional/ConfigurationTest.php
+++ b/modules/oe_webtools_analytics/tests/src/Functional/ConfigurationTest.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Drupal\Tests\oe_webtools_analytics\Functional;
 
 use Drupal\Tests\BrowserTestBase;
+use Drupal\Tests\oe_webtools\Traits\ContentAssertionTrait;
 
 /**
  * Tests that the configured settings are correctly output in the page.
@@ -12,6 +13,8 @@ use Drupal\Tests\BrowserTestBase;
  * @group oe_webtools_analytics
  */
 class ConfigurationTest extends BrowserTestBase {
+
+  use ContentAssertionTrait;
 
   /**
    * {@inheritdoc}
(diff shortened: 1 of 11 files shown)


modules/oe_webtools_cookie_consent/oe_webtools_cookie_consent.module
```diff
@@ -40,18 +41,13 @@ function oe_webtools_cookie_consent_page_attachments(array &$attachments) {
...
-    $cck_attachment = [
+
+    $page_bottom['oe_webtools_cookie_consent'][] = [
       '#type' => 'html_tag',
       '#tag' => 'script',
       '#value' => Json::encode($script_values),
       '#attributes' => ['type' => 'application/json'],
     ];
-
-    $event_cache->applyTo($attachments);
```
drishu, line 50: I removed this one entirely, I didn't see the point

brummbar (changes requested):
I don't see test changes nor failures, so probably we need to add tests to check that the tags are rendered inside the body.

modules/oe_webtools_analytics/tests/src/Functional/ConfigurationTest.php
```diff
@@ -42,13 +45,13 @@ public function testLibraryLoading(): void {
 
     $this->drupalLogout();
     $this->drupalGet('<front>');
-    $this->assertSession()->responseContains('<script type="application/json">{"utility":"piwik","siteID":"123","sitePath":["ec.europa.eu"],"instance":"testing"}</script>');
+    $this->assertBodyContainsApplicationJson('{"utility":"piwik","siteID":"123","sitePath":["ec.europa.eu"],"instance":"testing"}');
```
drishu, line 48: Here the front is the login page, which was accessed before, so if I dont add the event cache which contains the url.path context and the module settings tag, then this will fail here as it wont contain the json but what was cached in the previous login page call.

brummbar (approved):

PR #254, opened 2024-10-11
OEL-3178: Add maps version 3.0

drishu:
### Description

Add support for maps version 3

### Change log

- Added: Config form
- Changed: field widget

### DIFF
--- a/modules/oe_webtools_maps/config/schema/oe_webtools_maps.schema.yml
+++ b/modules/oe_webtools_maps/config/schema/oe_webtools_maps.schema.yml
@@ -4,7 +4,7 @@ oe_webtools_maps.settings:
   mapping:
     map_version:
       type: string
-      label: 'Map Version'
+      label: 'Map version'
 
 field.formatter.settings.oe_webtools_maps_map:
   type: mapping
--- a/modules/oe_webtools_maps/config/schema/oe_webtools_maps.schema.yml
+++ b/modules/oe_webtools_maps/config/schema/oe_webtools_maps.schema.yml
@@ -1,3 +1,11 @@
+oe_webtools_maps.settings:
+  type: config_object
+  label: 'Webtools Maps Settings'
+  mapping:
+    map_version:
+      type: string
+      label: 'Map Version'
+
 field.formatter.settings.oe_webtools_maps_map:
   type: mapping
   label: Webtools Map formatter settings
--- a/modules/oe_webtools_maps/config/schema/oe_webtools_maps.schema.yml
+++ b/modules/oe_webtools_maps/config/schema/oe_webtools_maps.schema.yml
@@ -1,6 +1,6 @@
 oe_webtools_maps.settings:
   type: config_object
-  label: 'Webtools Maps Settings'
+  label: 'Webtools Maps settings'
   mapping:
     map_version:
       type: string
--- a/modules/oe_webtools_maps/oe_webtools_maps.permissions.yml
+++ b/modules/oe_webtools_maps/oe_webtools_maps.permissions.yml
@@ -1,0 +1,2 @@
+administer webtools maps form:
+  title: 'Administer Webtools Maps Form'
--- a/modules/oe_webtools_maps/oe_webtools_maps.permissions.yml
+++ b/modules/oe_webtools_maps/oe_webtools_maps.permissions.yml
@@ -1,2 +1,2 @@
-administer webtools maps form:
-  title: 'Administer Webtools Maps Form'
+administer webtools maps:
+  title: 'Administer Webtools Maps'
--- a/modules/oe_webtools_maps/oe_webtools_maps.routing.yml
+++ b/modules/oe_webtools_maps/oe_webtools_maps.routing.yml
@@ -2,6 +2,6 @@ oe_webtools_maps.settings:
   path: '/admin/config/system/oe_webtools_maps'
   defaults:
     _form: 'Drupal\oe_webtools_maps\Form\WebtoolsMapsSettingsForm'
-    _title: 'Webtools Maps Form settings'
+    _title: 'Webtools Maps settings'
   requirements:
     _permission: 'administer webtools maps'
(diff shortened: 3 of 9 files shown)


modules/oe_webtools_maps/oe_webtools_maps.permissions.yml
```diff
@@ -0,0 +1,2 @@
+administer webtools maps form:
```
brummbar, line 1: This is not in line with the remaining permissions, it should be `administer webtools maps`
https://github.com/openeuropa/oe_webtools/blob/master/modules/oe_webtools_cookie_consent/oe_webtools_cookie_consent.permissions.yml#L1
https://github.com/openeuropa/oe_webtools/blob/master/modules/oe_webtools_globan/oe_webtools_globan.permissions.yml#L1
https://github.com/openeuropa/oe_webtools/blob/master/modules/oe_webtools_cookie_consent/oe_webtools_cookie_consent.permissions.yml#L1

modules/oe_webtools_maps/oe_webtools_maps.routing.yml
```diff
@@ -0,0 +1,7 @@
+oe_webtools_maps.settings:
+  path: '/admin/config/system/oe_webtools_maps'
+  defaults:
+    _form: 'Drupal\oe_webtools_maps\Form\WebtoolsMapsSettingsForm'
+    _title: 'Webtools Maps Form settings'
```
brummbar, line 5:
```suggestion
    _title: 'Webtools Maps settings'
```
Drop the form a bit everywhere :)

modules/oe_webtools_maps/config/schema/oe_webtools_maps.schema.yml
```diff
@@ -1,3 +1,11 @@
+oe_webtools_maps.settings:
+  type: config_object
+  label: 'Webtools Maps Settings'
+  mapping:
+    map_version:
+      type: string
+      label: 'Map Version'
```
brummbar, line 7:
```suggestion
      label: 'Map version'
```

brummbar (changes requested):
We don't need a new major, we can provide an update path that creates the configuration and sets it to 2.0 for existing installations. New installations can be on 3.0. Sites like EWPP can stick on 2.0 by exporting the configuration.

piotrsmykaj:
Hello @drishu,
Here is missing `oe_webtools_maps.links.menu.yml` and the settings form page is not visible on the Admin Configuration page.

modules/oe_webtools_maps/src/Form/WebtoolsMapsSettingsForm.php
```diff
@@ -0,0 +1,59 @@
...
+    $config = $this->config('oe_webtools_maps.settings');
+
+    $form['map_version'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Map Version'),
+      '#options' => [
+        '2.0' => $this->t('Version 2.0'),
+        '3.0' => $this->t('Version 3.0'),
+      ],
+      '#default_value' => $config->get('map_version') ?? '2.0',
```
drishu, line 35: @piotrsmykaj no need for "hook_post_update() and set version 2.0 for the existing websites." as we have this here

sergepavle:
Unfortunately, somehow, I overlooked this PR regarding the webtools map upgrade and worked on almost the same independently. See PR: https://github.com/openeuropa/oe_webtools/pull/258. 
It doesn't have BC but in my PR implemented some changes regarding webtools Map 3 API changes (missed in this PR).
Here https://github.com/openeuropa/oe_webtools/pull/258/files#diff-d5788593f82f730c0aae0cc4840492e4a812488fa1fb087015bec48283327268L90 you may notice that in new version UAC had a new structure like `layers.markers[0].data` instead of `layers[0].markers` (as worked before). Without that markers
will not be shown.
Maybe you could also extend test coverage by reusing these changes: https://github.com/openeuropa/oe_webtools/pull/258/files#diff-93e9dfce3433d92e9cae1a182d4fb153f8b8169475098c75580da33298e5a187R1 to make sure that formatted settings also was reflected in UAC webtools code.

modules/oe_webtools_maps/src/Plugin/Field/FieldFormatter/WebtoolsMapFormatter.php
```diff
@@ -71,11 +129,13 @@ public function settingsSummary() {
...
 
     foreach ($items as $delta => $item) {
       $data_array = [
         'service' => 'map',
-        'version' => '2.0',
+        // Fallback to version 2.0 for BC.
+        'version' => $config->get('map_version') ?? '2.0',
         'map' => [
           'zoom' => $this->getSetting('zoom_level'),
           'center' => [
```
sergepavle, line 141:
```
if ($this->getSetting('show_marker')) {
        $data_array['layers'] = [
```
should be updated according to changes in Webtools Maps UEC for maps 3.0. See this changes: https://github.com/openeuropa/oe_webtools/pull/258/files#diff-d5788593f82f730c0aae0cc4840492e4a812488fa1fb087015bec48283327268L90

modules/oe_webtools_maps/tests/src/Functional/ConfigurationTest.php
```diff
@@ -0,0 +1,128 @@
...
+      ])
+      ->save();
+
+    $this->drupalLogin($this->drupalCreateUser([], NULL, TRUE));
+  }
+
+  /**
+   * Tests if changing configuration changes the map JSON.
+   */
+  public function testConfigurationChanges(): void {
```
sergepavle, line 85: As we didn't have at all test coverage for oe_webtools_maps probably it makes sense to extend test coverage to assert field formatter settings. See https://github.com/openeuropa/oe_webtools/pull/258/files#diff-93e9dfce3433d92e9cae1a182d4fb153f8b8169475098c75580da33298e5a187R71

modules/oe_webtools_maps/src/Plugin/Field/FieldFormatter/WebtoolsMapFormatter.php
saidatom: I confirm that, without this change, the marker does not appear on the map.

saidatom:
An additional change following the update to Maps 3.0 is located in the file **tests/Behat/WebtoolsMapsContext.php**
Example:
```
-            if ($map_data->layers[0]->markers->features[0]->properties->name !== $value) {
+            if ($map_data->layers->markers[0]->data->features[0]->properties->name !== $value) {
```

sergepavle (approved):

PR #260, opened 2025-01-27
OEL-3585: eTrans unified block.

Aporie:
### Description

Add a block that allows to unify the UX for page translation requests using the drupal language switcher and webtool eTrans service.


modules/oe_webtools_etrans/tests/src/FunctionalJavascript/UnifiedEtransBlockTest.php
```diff
@@ -0,0 +1,115 @@
...
+    $this->assertSession()->pageTextContains("Je suis un texte qui va être traduit.");
+    $this->assertSession()->pageTextNotContains("French is available via eTranslation, the European Commission's machine translation service.");
+    // Croatian translation should have the block
+    // and display the english text.
+    $this->drupalGet($node->toUrl(NULL, [
+      'language' => ConfigurableLanguage::load('hr'),
+    ]));
+    $this->assertSession()->pageTextContains("I'm a text that will be translated.");
+    $this->assertSession()->pageTextContains("Croatian is available via eTranslation, the European Commission's machine translation service.");
+
```
drishu, line 99: the whole visibility and json of the block depends on the entity. So at this point, lets simulate a translation being added (programatically) and lets see if it gets picked up without flushing.

drishu (changes requested):

modules/oe_webtools_etrans/tests/src/FunctionalJavascript/UnifiedEtransBlockTest.php
Aporie: Done

PR #262, opened 2025-01-28
OEL-3585: eTrans unified block.

Aporie:
### Description

Add a block that allows to unify the UX for page translation requests using the drupal language switcher and webtool eTrans service.


modules/oe_webtools_etrans/tests/src/FunctionalJavascript/UnifiedEtransBlockTest.php
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  protected $defaultTheme = 'stark';
+
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'block',
```
drishu, line 24: enable page_cache and dynamic_page_cache to be sure caching is enabled

drishu (changes requested):
also branch needs a rebase with master

modules/oe_webtools_etrans/README.md
```diff
@@ -0,0 +1,29 @@
+// phpcs:ignoreFile
```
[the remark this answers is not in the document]
Aporie: Updated.

modules/oe_webtools_etrans/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+# OpenEuropa Webtools eTrans
+
+This component provides the integration with the Webtools eTranslation service.
```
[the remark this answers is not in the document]
Aporie: Updated

modules/oe_webtools_etrans/tests/src/FunctionalJavascript/UnifiedEtransBlockTest.php
Aporie: Updated.

drishu (approved):

drishu (approved):

upchuk (changes requested):

modules/oe_webtools_etrans/js/oe_webtools_etrans_unified.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+(function (Drupal) {
+
+  /**
+   * Handles the machine translation.
+   *
+   * @type {Drupal~behavior}
+   */
+  Drupal.behaviors.etrans = {
```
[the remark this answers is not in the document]
Aporie: Thanks. Updated.

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
```diff
@@ -0,0 +1,309 @@
...
+   * Default values for the configuration options of this block.
+   */
+  protected const DEFAULT_CONFIGURATION = [
+    'delay' => 0,
+    'include' => '',
+    'exclude' => '',
+  ];
+
+  /**
+   * Creates an ETransBlock.
```
nagyad, line 43: Creates an UnifiedEtransBlock.

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
```diff
@@ -0,0 +1,309 @@
...
+      $plugin_definition,
+      $container->get('language_manager'),
+      $container->get('current_route_match')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function access(AccountInterface $account, $return_as_object = FALSE) {
```
nagyad, line 82: Blocks with specific access checking should override blockAccess() method rather than access(), in order to avoid repeating the handling of the $return_as_object argument. Please see BlockPluginTrait.

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
```diff
@@ -0,0 +1,309 @@
...
+    $translation_exists = (bool) $this->getRouteEntityLangcode();
+    return AccessResult::allowedIf(!$translation_exists);
```
nagyad, line 84: We should apply route cache context.

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
```diff
@@ -0,0 +1,309 @@
...
+      '#title' => $this->t('Important information about machine translation', [], ['langcode' => $translation_to_language->getId()]),
+    ];
+    $build['translation_request'] = [
+      '#theme' => 'block__unified_etrans_request',
+      '#message' => $message,
+      '#translate_link' => $translate_link,
+      '#disclaimer_link' => $disclaimer_link,
+    ];
+
+    return $build;
```
nagyad, line 167: I think the translation_to_language should be added as a cacheable dependency.

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
```diff
@@ -0,0 +1,309 @@
...
+   *
+   * @return string
+   *   Requested langcode, default one or empty string.
+   */
+  protected function getRouteEntityLangcode(bool $default = FALSE): string {
+    $route = $this->routeMatch->getRouteObject();
+    $requested_langcode = $this->languageManager->getCurrentLanguage()->getId();
+    if ($route->hasRequirement('node')) {
+      /** @var \Drupal\node\NodeInterface $entity */
+      $entity = $this->routeMatch->getParameter('node');
```
nagyad, line 235: Are we sure that entity is an instance of NodeInterface at this point?

nagyad (changes requested):

modules/oe_webtools_etrans/oe_webtools_etrans.libraries.yml
```diff
@@ -0,0 +1,11 @@
+wt_etrans:
+  version: VERSION
```
nagyad, line 2: Why do we need this here?

nagyad (changes requested):

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
Aporie: Please provide more info or doc, I do not understand.

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
Aporie: Well, you need the block to work for anonymous user. The language is selected from the language switcher. As cache is really hard to implement for anonymous user, I think it's an overkill and really hard to implement to actually cache the $translate_to_language var in the block.

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
Aporie: Except if someone implements {node} parameter as expecting a taxonomy term in their route, we should expect a NodeInterface yes. Anyway, this code is only informative and does not impact its computing. Now, most of the function called later on are actually depending on TranslatableEntityInterface, so renaming the variable $node and keeping NodeInterface, or keeping $entity and indicating TranslatableEntityInterface ... At the end, it'll all depend on what other modules send in their routes with the parameter {node}, which we can't predict.

modules/oe_webtools_etrans/oe_webtools_etrans.libraries.yml
Aporie: If I recall correctly, this allows the js to be rebuilt on drupal updates. I personally have a habit to put this.

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
Aporie: Thanks, updated.

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
upchuk: Recently I learned that checkAccess gets called regardless, so it doesn't really matter.

modules/oe_webtools_etrans/oe_webtools_etrans.libraries.yml
nagyad: It is used as an identifier for your libraries. VERSION placeholder will just make \Drupal::VERSION, which is kind of pointless unless this is core module's library. Also, we had an issue with this placeholder used in a contrib: https://www.drupal.org/project/maxlength/issues/3326716

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
```diff
@@ -79,7 +79,7 @@ public static function create(ContainerInterface $container, array $configuratio
   /**
    * {@inheritdoc}
    */
-  public function access(AccountInterface $account, $return_as_object = FALSE) {
+  public function blockAccess(AccountInterface $account, $return_as_object = FALSE) {
```
nagyad, line 82: The blockAccess() has no second parameter. See: BlockPluginTrait::blockAccess()

nagyad (changes requested):

modules/oe_webtools_etrans/oe_webtools_etrans.libraries.yml
Aporie: That's interesting because I see related issues complaining about exactly the opposite of what this version var is supposed to do: https://www.drupal.org/project/klaro/issues/3497609. I'll remove it, but there might be further digging to do to get to the bottom of that.

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
Aporie: Of course, if I'm checking access(). Updated.

nagyad (approved):
Thank you for this contribution!

Issue #271, opened 2025-03-04
Oe_webtools_unified_etrans does not work for non ISO language format


Aporie:
When using the oe_webtools_etrans block, and browsing the website in Portuguese for example, the webtool translation widget will not work when trying to translate the page.

This is due to Drupal using browser langcode format and not ISO 639-2 as expected by webtool.

PR incoming.

vever001:
Interesting, I also noticed ETransBlock.php:95 which uses
`substr($this->languageManager->getCurrentLanguage()->getId(), 0, 2),`

This looks wrong and we could probably leverage the `language.mappings` config from core's language module.
Probably via `language_get_browser_drupal_langcode_mappings()`.

Could we expand the scope to include this?

Aporie:
I wasn't aware of this function `language_get_browser_drupal_langcode_mappings`, it seems to do exactly what I was looking for.
I did insert a [dependency to drupal:language](https://github.com/openeuropa/oe_webtools/blob/master/modules/oe_webtools_etrans/oe_webtools_etrans.info.yml#L7) for webtools_etrans last time, so I guess we can just safely use it.

Regarding eTrans and eTransUnified block, there might be some merging to do here, as you mentioned. We could either create a base block, extended by the two, or a service to map to ISO 639-2.

Don't think `language_get_browser_drupal_langcode_mappings` will be enough anyway, in case custom language implementations are used (which shouldn't be the case I guess, but who knows ...).

Aporie:
> I am not a maintainer of oe_webtools, but I don't think a base class is justified. I would just use that function in each class, it's not much, in case of ETransBlock:

Neither am I. I only contributed to eTransUnified block.

`Why not? this is a config so any website can change this.`

That is why you need some kind of fallback when the mapping does not exist. `$excluded_langcode = $language_mappings[$langcode] ?? $langcode;` is too simple and induces terrible UX. When trying to translate a content, webtools will return nothing, making the "translate" button appears like it's not doing anything.

`But I didn't really look into the eTransUnified block. A trait could also be another option.`

A trait is an option, but those two blocks share more code than just this one need to retrieve potential ISO 639-2 code from drupal language config.

Anyway, I've updated my code with your function. Thanks for sharing this.

donquixote:
Another problem with the function is that it does not support dependency injection.

PR #272, opened 2025-03-04
D8TEFD-1822: Fix drupal non ISO langcode.

Aporie:
### Description

Issue #271

### Change log

- Added: mapLangcodeToIso, a function that attempts to get drupal langcode in ISO 639-2 expected by webtools.

### DIFF
--- a/modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
+++ b/modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
@@ -312,13 +312,22 @@ class UnifiedEtransBlock extends BlockBase implements ContainerFactoryPluginInte
   }
 
   /**
-   * Attempt to map langcode to ISO 639-1.
+   * Attempts to map a Drupal langcode to ISO 639-1 as expected by Webtools.
+   *
+   * Language codes in Drupal are based on w3c language tags.
+   * For most (European) languages, these are identical with the ISO-639-1
+   * code.
+   * In some cases, the language code in Drupal has an additional suffix to
+   * distinguish regional variants, e.g. 'pt-pt' vs 'pt-br'.
+   * In other cases, Drupal uses a different ISO 639-1 code than Webtools
+   * does: For Norwegian, Drupal has 'nb' for Bokmal, whereas Webtools expects
+   * the more generic 'no' for Norwegian.
    *
    * @param string $langcode
    *   The Drupal langcode.
    *
    * @return string
-   *   The ISO 639-1 langcode format.
+   *   The corresponding ISO 639-1 language code as expected by Webtools.
    */
   protected function mapLangcodeToIso(string $langcode): string {
     // In Drupal, langcode are often in ISO 639-1 format by default.
--- a/modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
+++ b/modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Drupal\oe_webtools_etrans\Plugin\Block;
 
 use Drupal\Component\Serialization\Json;
-use Drupal\Component\Serialization\Yaml;
 use Drupal\Component\Utility\Html;
 use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Block\Attribute\Block;
@@ -325,9 +324,7 @@ class UnifiedEtransBlock extends BlockBase implements ContainerFactoryPluginInte
     // In Drupal, langcode are often in ISO 639-2 format by default.
     // But, some languages use browser formats like 'pt-pt' or 'ta-lk'.
     // A mapping exists in the language module.
-    $mapping_file = file_get_contents(DRUPAL_ROOT . '/core/modules/language/config/install/language.mappings.yml');
-    $mapping = Yaml::decode($mapping_file ?: '');
-    $mapping = array_flip($mapping['map'] ?? []);
+    $mapping = array_flip(language_get_browser_drupal_langcode_mappings());
     if (isset($mapping[$langcode])) {
       return substr($mapping[$langcode], 0, 2);
     }
--- a/modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
+++ b/modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
@@ -153,7 +153,7 @@ class UnifiedEtransBlock extends BlockBase implements ContainerFactoryPluginInte
     ];
     $disclaimer_link = [
       '#type' => 'link',
-      '#url' => Url::fromUri('https://commission.europa.eu/languages-our-websites/use-machine-translation-europa_' . $current_language->getId(), [
+      '#url' => Url::fromUri('https://commission.europa.eu/languages-our-websites/use-machine-translation-europa_' . $langcode_to, [
         'attributes' => [
           'class' => ['webtools-etrans--disclaimer'],
           'target' => '_blank',
@@ -312,16 +312,16 @@ class UnifiedEtransBlock extends BlockBase implements ContainerFactoryPluginInte
   }
 
   /**
-   * Attempt to map langcode to ISO 639-2.
+   * Attempt to map langcode to ISO 639-1.
    *
    * @param string $langcode
    *   The drupal langcode.
    *
    * @return string
-   *   The ISO 639-2 langcode format.
+   *   The ISO 639-1 langcode format.
    */
   protected function mapLangcodeToIso(string $langcode): string {
-    // In Drupal, langcode are often in ISO 639-2 format by default.
+    // In Drupal, langcode are often in ISO 639-1 format by default.
     // But, some languages use browser formats like 'pt-pt' or 'ta-lk'.
     // A mapping exists in the language module.
     $mapping = array_flip(language_get_browser_drupal_langcode_mappings());
--- a/modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
+++ b/modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
@@ -333,6 +333,8 @@ class UnifiedEtransBlock extends BlockBase implements ContainerFactoryPluginInte
     // Get the browser language lookup map from language module.
     // This maps different alternative language codes to their corresponding
     // Drupal language codes.
+    // @todo The browser language map is not really meant for this purpose.
+    //   It just happens to do what we need in the most common scenarios.
     $mappings = language_get_browser_drupal_langcode_mappings();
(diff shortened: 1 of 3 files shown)


modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
```diff
@@ -134,28 +138,28 @@ public function build(): array {
...
+        '@language' => $current_language->getName(),
       ], [
-        'langcode' => $translation_to_language->getId(),
+        'langcode' => $current_language->getId(),
       ]),
     ];
     $disclaimer_link = [
       '#type' => 'link',
-      '#url' => Url::fromUri('https://commission.europa.eu/languages-our-websites/use-machine-translation-europa_' . $translation_to_language->getId(), [
+      '#url' => Url::fromUri('https://commission.europa.eu/languages-our-websites/use-machine-translation-europa_' . $current_language->getId(), [
```
piotrsmykaj, line 156:
```suggestion
      '#url' => Url::fromUri('https://commission.europa.eu/languages-our-websites/use-machine-translation-europa_' . $langcode_to, [
```

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
```diff
@@ -134,28 +138,28 @@ public function build(): array {
...
     ]);
     $translate_link = [
       '#type' => 'link',
       '#url' => Url::fromRoute('<none>', [], ['attributes' => ['class' => ['oe-webtools-unified-etrans--translate']]]),
       '#title' => $this->t('Translate to @language', [
-        '@language' => $translation_to_language->getName(),
+        '@language' => $current_language->getName(),
       ], [
-        'langcode' => $translation_to_language->getId(),
+        'langcode' => $current_language->getId(),
```
piotrsmykaj, line 151:
```suggestion
        'langcode' => $langcode_to,
```

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
```diff
@@ -134,28 +138,28 @@ public function build(): array {
...
       '#type' => 'link',
-      '#url' => Url::fromUri('https://commission.europa.eu/languages-our-websites/use-machine-translation-europa_' . $translation_to_language->getId(), [
+      '#url' => Url::fromUri('https://commission.europa.eu/languages-our-websites/use-machine-translation-europa_' . $current_language->getId(), [
         'attributes' => [
           'class' => ['webtools-etrans--disclaimer'],
           'target' => '_blank',
         ],
       ]),
-      '#title' => $this->t('Important information about machine translation', [], ['langcode' => $translation_to_language->getId()]),
+      '#title' => $this->t('Important information about machine translation', [], ['langcode' => $current_language->getId()]),
```
piotrsmykaj, line 162:
```suggestion
      '#title' => $this->t('Important information about machine translation', [], ['langcode' => $langcode_to]),
```

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
```diff
@@ -306,4 +311,27 @@ protected function preparesWtEtransJson(string $translation_from, string $placeh
     return $json;
   }
 
+  /**
+   * Attempt to map langcode to ISO 639-2.
```
piotrsmykaj, line 315:
```suggestion
   * Attempt to map langcode to Drupal language codes.
```

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
```diff
@@ -306,4 +311,27 @@ protected function preparesWtEtransJson(string $translation_from, string $placeh
...
+   *
+   * @param string $langcode
+   *   The drupal langcode.
+   *
+   * @return string
+   *   The ISO 639-2 langcode format.
+   */
+  protected function mapLangcodeToIso(string $langcode): string {
+    // In Drupal, langcode are often in ISO 639-2 format by default.
```
piotrsmykaj, line 324: ISO 639-1 -> two-letters codes
ISO 639-2 -> three-letters codes

piotrsmykaj, line 324:
```suggestion
    // In Drupal, langcode are often in ISO 639-1 format by default.
```
ISO 639-1 -> two-letters codes
ISO 639-2 -> three-letters codes

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
Aporie: No, it does map to ISO-639-1 (you're correct below. alpha-2 is 3 letters, alpha-1 is 2).

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
Aporie: Correct. We can commit this.

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
Aporie, line 315:
```suggestion
   * Attempt to map langcode to ISO 639-1.
```

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
```diff
@@ -306,4 +311,27 @@ protected function preparesWtEtransJson(string $translation_from, string $placeh
...
   }
 
+  /**
+   * Attempt to map langcode to ISO 639-1.
+   *
+   * @param string $langcode
+   *   The drupal langcode.
+   *
+   * @return string
+   *   The ISO 639-2 langcode format.
```
piotrsmykaj, line 321:
```suggestion
   *   The ISO 639-1 langcode format.
```

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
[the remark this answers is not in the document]
piotrsmykaj: it's oke

piotrsmykaj (approved):

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
```diff
@@ -101,10 +101,14 @@ public function build(): array {
...
-    $translation_to_language = $this->languageManager->getCurrentLanguage();
+    $current_language = $this->languageManager->getCurrentLanguage();
+    $langcode_from = $this->getRouteEntityLangcode(TRUE) ?: $this->languageManager->getDefaultLanguage()->getId();
+    $langcode_to = $current_language->getId();
 
-    $json = $this->preparesWtEtransJson($translation_from, $placeholder_id);
+    $langcode_from = $this->mapLangcodeToIso($langcode_from);
+    $langcode_to = $this->mapLangcodeToIso($langcode_to);
+
+    $json = $this->preparesWtEtransJson($langcode_from, $placeholder_id);
```
donquixote, line 111: I think we should use different variable names for the ISO langcodes, I think it would be more clear.
```suggestion
    $iso_langcode_from = $this->mapLangcodeToIso($langcode_from);
    $iso_langcode_to = $this->mapLangcodeToIso($langcode_to);

    $json = $this->preparesWtEtransJson($iso_langcode_from, $placeholder_id);
```

donquixote: Of course the idea would be to use two distinct variable names:
- The Drupal langcode is `$langcode_from`.
- The ISO langcode is `$iso_langcode_from`.

So yes, we need to apply the rename further down, as you did (I did not want to go crazy with this suggestion).
But before the `->mapLangcodeToIso()` we want to keep the old name.

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
```diff
@@ -306,4 +311,27 @@ protected function preparesWtEtransJson(string $translation_from, string $placeh
     return $json;
   }
 
+  /**
+   * Attempt to map langcode to ISO 639-1.
+   *
+   * @param string $langcode
+   *   The drupal langcode.
```
donquixote, line 318:
```suggestion
   *   The Drupal langcode.
```

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
Aporie: @donquixote, let's not block this PR for such details.
If really that doesn't suit you, feel free to commit your changes.

donquixote: I am happy to do that! Thanks for the green light.

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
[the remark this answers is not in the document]
donquixote: Yep it should be Drupal language id.

modules/oe_webtools_etrans/src/Plugin/Block/UnifiedEtransBlock.php
```diff
@@ -306,4 +311,27 @@ protected function preparesWtEtransJson(string $translation_from, string $placeh
...
+   *   The Drupal langcode.
+   *
+   * @return string
+   *   The ISO 639-1 langcode format.
+   */
+  protected function mapLangcodeToIso(string $langcode): string {
+    // In Drupal, langcode are often in ISO 639-1 format by default.
+    // But, some languages use browser formats like 'pt-pt' or 'ta-lk'.
+    // A mapping exists in the language module.
+    $mapping = array_flip(language_get_browser_drupal_langcode_mappings());
```
donquixote: One problem with this function is that it does not support dependency injection.
It might be cleaner to just use the config.
The only thing this function does differently is checking for `$config->isNew()`. The function was added in https://www.drupal.org/node/365615 in 2012. In this commit it was also used in a form, and this is probably the reason why they call `$config->isNew()`.
For us there is probably no reason why we would need this check. So it would be ok if we simply use the config.
The only other place where this is called in core is in ckeditor5.

donquixote (commented):
I pushed some changes, which are TBD - we can drop or revert commits if we disagree.

I also added a bunch of comments, some of them will now appear as outdated due to the changes I pushed.

Using `language_get_browser_drupal_langcode_mappings()` is risky. On most sites it will do what we want, for those languages that we care about, but this is more by coincidence than by design. A more future-proof solution would be a dedicated lookup map.

piotrsmykaj (approved):

PR #284, opened 2025-07-07
EWPP-5701: Webtools search widget.

hernani:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


intelektron (approved):

modules/oe_webtools_search_widget/src/Plugin/Block/SearchWidgetBlock.php
```diff
@@ -0,0 +1,142 @@
...
+   * {@inheritdoc}
+   */
+  protected function blockAccess(AccountInterface $account) {
+    return AccessResult::allowedIfHasPermission($account, 'access content');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
```
22Alexandra, line 48: Just a suggestion, we could add `defaultConfiguration` with default empty for providing a form selector to be used in the build with the webtools json, since they allow this now.

PR #313, opened 2026-04-10
EWPP-6706: Create a new field type and widget that use the Webtools PickIcons service.

22Alexandra:


modules/oe_webtools_wepick_icons/src/Element/WePickIcons.php
```diff
@@ -0,0 +1,132 @@
...
+    $element['#attributes']['maxlength'] = '100000000';
+    $element['#maxlength'] = '100000000';
+
+    return $element;
+  }
+
+  /**
+   * Form element validation handler for WePick Icons elements.
+   */
+  public static function validateWepickIcons(array &$element, FormStateInterface $form_state, array &$complete_form) {
```
nagyad: Method name mismatch (validateWePickIcons vs validateWepickIcons). Let's make sure we are following the same pattern.

modules/oe_webtools_wepick_icons/config/schema/oe_webtools_wepick_icons.schema.yml
```diff
@@ -0,0 +1,25 @@
+field.widget.settings.oe_webtools_wepick_icons:
```
nagyad, line 12: Missing field.value.oe_webtools_wepick_icons config schema type

modules/oe_webtools_wepick_icons/src/Plugin/Field/FieldWidget/WePickIconsWidget.php
```diff
@@ -0,0 +1,170 @@
...
+    ];
+
+    foreach ($map as $setting_suffix => $json_key) {
+      $value = $this->getSetting("{$type}_{$setting_suffix}");
+      if (!empty($value)) {
+        $filter[$json_key] = array_map('trim', explode(',', $value));
+      }
+    }
+
+    return $filter;
```
nagyad, line 167: If a setting contains only commas/spaces (e.g. " , , "), the result is ["", "", ""] passed to the Webtools API. Fix: array_filter(array_map('trim', explode(',', $value))). Unlikely input but easy defensive fix.

nagyad:
Good work.

nagyad (changes requested):
Please check my comments for consideration.

nagyad (approved):

modules/oe_webtools_wepick_icons/tests/src/Functional/WePickIconsWidgetTest.php
```diff
@@ -0,0 +1,153 @@
...
+use Drupal\Core\Entity\Entity\EntityFormDisplay;
+use Drupal\Tests\BrowserTestBase;
+use Drupal\field\Entity\FieldConfig;
+use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\node\Entity\Node;
+
+/**
+ * Tests the WePick Icons widget using the WePick Icons form element.
+ */
+class WePickIconsWidgetTest extends BrowserTestBase {
```
intelektron, line 16: The widget is ok, but don't we need some simple formatter to make it usable? 

Right now, the field can only be disabled.

22Alexandra: I thought about it but other than simply rendering the icon name and icon family it would require the rendering from oe_theme. So, I decided that anyone using this service should handle the rendering in their project.

intelektron (approved):

